PR #2, opened 2020-08-04
Scaffold: 2 – Add sentry and simple provider utils

andy-hook:
Looked at supporting dynamic updates to the `environment` event prop for future runtime config but it wasn't super straightforward nor totally necessary at this exact moment.

### DIFF
--- a/src/environment.js
+++ b/src/environment.js
@@ -1,16 +1,13 @@
-// Fallback to rinkeby
-const FALLBACK_CHAIN_ID = 4
+// Rinkeby
+const DEFAULT_CHAIN_ID = 4
 
 const ENV_VARS = {
   BUILD() {
     return process.env.BUILD || 'undefined'
   },
-  DEFAULT_CHAIN_ID() {
+  CHAIN_ID() {
     const chainId = parseInt(process.env.CHAIN_ID)
-    return isNaN(chainId) ? FALLBACK_CHAIN_ID : chainId
-  },
-  ENABLE_SENTRY() {
-    return process.env.ENABLE_SENTRY === '1'
+    return isNaN(chainId) ? DEFAULT_CHAIN_ID : chainId
   },
   FORTMATIC_API_KEY() {
     return process.env.FORTMATIC_API_KEY || ''
--- a/src/lib/web3-utils.js
+++ b/src/lib/web3-utils.js
@@ -1,0 +1,8 @@
+export function getNetworkName(chainId) {
+  chainId = String(chainId)
+
+  if (chainId === '1') return 'Ethereum'
+  if (chainId === '4') return 'Rinkeby'
+
+  return 'Unknown'
+}
--- a/src/sentry.js
+++ b/src/sentry.js
@@ -1,0 +1,15 @@
+import { init as initSentry } from '@sentry/browser'
+import env from './environment'
+import { getNetworkName } from './lib/web3-utils'
+
+export const sentryEnabled = Boolean(env('SENTRY_DSN'))
+
+export default function initializeSentry() {
+  if (sentryEnabled) {
+    initSentry({
+      dsn: env('SENTRY_DSN'),
+      environment: getNetworkName(env('CHAIN_ID')),
+      release: 'network-dashboard@' + env('BUILD'),
+    })
+  }
+}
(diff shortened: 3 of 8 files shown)


src/lib/web3-utils.js
```diff
@@ -0,0 +1,39 @@
+import env from '../environment'
+
+export function getUseWalletProviders() {
+  const providers = [{ id: 'injected' }, { id: 'frame' }]
+
+  if (env('FORTMATIC_API_KEY')) {
+    providers.push({
+      id: 'fortmatic',
+      useWalletConf: { apiKey: env('FORTMATIC_API_KEY') },
+    })
```
bpierre, line 10: Wondering if the idea is to add more things here, or if we should define a useWallet()-compatible object directly?

src/sentry.js
```diff
@@ -0,0 +1,16 @@
+import { init as initSentry } from '@sentry/browser'
+import env from './environment'
+import { getNetworkName } from './lib/web3-utils'
+
+const SENTRY_DSN = env('SENTRY_DSN')
+export const sentryEnabled = Boolean(SENTRY_DSN && env('ENABLE_SENTRY'))
```
bpierre, line 6: Should we rely on the presence of `SENTRY_DSN`? Or do you see a reason to have `ENABLE_SENTRY` as well?

andy-hook: That's a good point, It was to prevent sending events during local development but in that instance the `SENTRY_DSN` isn't available anyway.

src/lib/web3-utils.js
andy-hook: There's more to come here, I think i'll remove this for now and add as part of a separate PR for the wallet

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "semi": false,
-  "singleQuote": true
+  "singleQuote": true,
+  "arrowParens": "avoid"
```
sohkai, line 4: Ah, I think we'll continue with using the default of this setting and make any changes required as we update prettier in other repos.

src/environment.js
```diff
@@ -19,7 +19,7 @@ const ENV_VARS = {
     return process.env.PORTIS_DAPP_ID || ''
   },
   SENTRY_DSN() {
-    const dsn = process.env.SENTRY_DSN || ''
+    const dsn = process.env.NEXT_PUBLIC_SENTRY_DSN || ''
```
sohkai, line 22: Right, let's actually rollback to explicitly providing the environment variable in the config and avoid using the Vercel frontend for now to do this!

Fleek won't have access to these variables.

src/lib/web3-utils.js
```diff
@@ -0,0 +1,39 @@
...
+      connectors[provider.id] = provider.useWalletConf
+    }
+    return connectors
+  }, {})
+}
+
+export function getNetworkName(chainId = env('CHAIN_ID')) {
+  chainId = String(chainId)
+
+  if (chainId === '1') return 'Mainnet'
```
sohkai, line 35:
```suggestion
  if (chainId === '1') return 'Ethereum'
```

Or alternatively "Ethereum mainnet" :). There are lots of "mainnets" around now, so it's better to be more explicit!

src/lib/web3-utils.js
```diff
@@ -0,0 +1,39 @@
...
+export function getUseWalletConnectors() {
+  return getUseWalletProviders().reduce((connectors, provider) => {
+    if (provider.useWalletConf) {
+      connectors[provider.id] = provider.useWalletConf
+    }
+    return connectors
+  }, {})
+}
+
+export function getNetworkName(chainId = env('CHAIN_ID')) {
```
sohkai, line 32: I find it kind of awkward to default the argument to the current environment, as a utility it might be better to be explicit about always requiring it.

.prettierrc
andy-hook: That was the previous default, it changed with the upgrade, I added it to match the other repos

sohkai: Yes exactly; it went undocumented but we made a decision a while back on prettier v2 to align with the new defaults :).

andy-hook: Oh, yes ok!

PR #4, opened 2020-08-05
Fix duplicated global style blocks

andy-hook:
The upgrade to styled-components v5 was causing duplicate global style blocks to be rendered in the head. They seemed to originate from `<Main/>` in aragon/ui. For now I think it's best to pin it to v4.

@delfipolito this should resolve the issue we discussed with HMR reloads.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
+import { Main } from '@aragon/ui'
 import App from './App'
 import initializeSentry from './sentry'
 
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -41,7 +41,7 @@
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-spring": "^8.0.1",
-    "styled-components": "^5.1.1"
+    "styled-components": "4.4.1"
```
sohkai, line 44: Perhaps we can still leave this with a `^` for the 4.x range

bpierre:
> Looks good; cc @bpierre perhaps we should set the peer dependency range on aragonUI to be `^4` instead of `>= 4`?

We could do that yes, but I suspect it will break apps already already using aragonUI with styled-components v5.

I would like to understand the issue as aragonUI should work with v5. @andy-hook could you tell me how to reproduce? It seems to work fine with the devbox or the client. :thinking:

andy-hook:
> > Looks good; cc @bpierre perhaps we should set the peer dependency range on aragonUI to be `^4` instead of `>= 4`?
> 
> We could do that yes, but I suspect it will break apps already already using aragonUI with styled-components v5.
> 
> I would like to understand the issue as aragonUI should work with v5. @andy-hook could you tell me how to reproduce? It seems to work fine with the devbox or the client. 🤔
> 
> Edit: in any case, it’s probably a good idea to use `>= 4 < 6` in case v6 breaks compatibility.

I think I've identified the issue with fresh eyes today. See https://github.com/styled-components/styled-components/issues/3008

PR #5, opened 2020-08-05
Scaffold: Add super simple routes

andy-hook:
Define the most basic set of initial routes we need. Error and route boundaries to come later.

### DIFF
--- a/src/components/Header/Header.js
+++ b/src/components/Header/Header.js
@@ -1,0 +1,20 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+import { GU } from '@aragon/ui'
+
+function Header() {
+  return (
+    <header
+      css={`
+        display: inline-grid;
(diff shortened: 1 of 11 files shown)


src/components/Header/Header.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
```
andy-hook, line 2: I'm using `Link` from the routing package here just as an example, we will likely end up using `useHistory` to augment our own components.

src/theme.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  _name: 'network-dashboard',
+  _appearance: 'light',
+}
```
andy-hook, line 4: This could be an easy way to use the new primary brand colours without having to update aragon/ui immediately.

PR #7, opened 2020-08-06
Disputable votes with Connect

delfipolito:
To try it out, link connect-disputable-voting locally.


src/components/Proposals/Proposals.js
```diff
@@ -1,15 +1,36 @@
...
+import ProposalCard from './ProposalCard'
+
+const useDisputableVoting = createAppHook(connectVoting)
 
 function Proposals() {
+  const [disputableVoting] = useApp('disputable-voting')
+  const [votes] = useDisputableVoting(disputableVoting, (app) => app.votes())
+  const { layoutName } = useLayout()
+  const compactMode = layoutName === 'small'
+  const rowHeight = compactMode ? null : 294
```
andy-hook, line 14: How do you feel about setting this to always be `294` for now and removing `useLayout`? The layout context doesn't exist yet so I think this would be equivalent anyway. We can come back and tweak it after!

src/components/Proposals/Proposals.js
```diff
@@ -1,15 +1,36 @@
...
-        <Link to="/proposals/1">Proposal 1</Link>
-        <Link to="/proposals/2">Proposal 2</Link>
-        <Link to="/proposals/3">Proposal 3</Link>
-      </div>
+      <section
+        css={`
+          width: 80vw;
+          margin: auto;
+        `}
+      >
```
andy-hook, line 23: Just a heads up that we will probably be limiting our layout in the `MainView` so we can remove this. It's not a big deal though and we can do it in a separate  improvement.

src/components/Proposals/Proposals.js
```diff
@@ -1,15 +1,36 @@
...
+import { CardLayout, GU, useLayout } from '@aragon/ui'
+import { createAppHook, useApp } from '@aragon/connect-react'
+import connectVoting from '@aragon/connect-voting-disputable'
+import ProposalCard from './ProposalCard'
+
+const useDisputableVoting = createAppHook(connectVoting)
 
 function Proposals() {
+  const [disputableVoting] = useApp('disputable-voting')
+  const [votes] = useDisputableVoting(disputableVoting, (app) => app.votes())
```
andy-hook, line 11: I think this works for now, but for organisation we will likely want to separate this into a hook `useDisputableVotes` or something :)

src/components/Proposals/ProposalCard.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Card, GU } from '@aragon/ui'
+import { useHistory } from 'react-router-dom'
```
andy-hook, line 4: nitpick: router import can come before aragon/ui :)

src/components/Proposals/ProposalCard.js
```diff
@@ -0,0 +1,31 @@
...
+
+function ProposalCard({ vote }) {
+  const history = useHistory()
+
+  return (
+    <Card
+      onClick={() => {
+        history.push(`/proposals/${vote.voteId}`)
+      }}
```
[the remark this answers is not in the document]
delfipolito: I move it to app-logic.js. Let me know what you think :)

src/components/Proposals/Proposals.js
delfipolito: I added a TODO, to modify it once we have styles on MainView

src/components/Proposals/Proposals.js
delfipolito: yes! I also added it to app-logic (to have it all there)

src/components/Proposals/ProposalCard.js
```diff
@@ -0,0 +1,31 @@
...
+      css={`
+        display: grid;
+        grid-template-columns: 100%;
+        grid-template-rows: auto 1fr auto auto;
+        grid-gap: ${1 * GU}px;
+        padding: ${3 * GU}px;
+      `}
```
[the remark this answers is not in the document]
delfipolito: Not for now. I'll add it later when we have more infoo to display there

src/app-logic.js
```diff
@@ -0,0 +1,39 @@
...
+  return votes
+}
+
+// Handles the main logic of the app.
+export function useAppLogic() {
+  const [selectedProposal, selectProposal] = useSelectedProposal()
+  return {
+    selectProposal,
+    selectedProposal,
+    getVotes: useGetVotes(),
```
sohkai, line 37: Wasn't sure, but does `useGetVotes()` return a function? It looks like it just gets the votes, so maybe we should call this `votes`?

Was also curious for the future if we could get the loading state; does `useDisputableVoting()` provide that to us?

delfipolito: I think in the future we can get the loading state with appsStatus.loading from useApps()

delfipolito: I will take care of this in the next PR

PR #8, opened 2020-08-07
Add simple network config with Connect provider

andy-hook:
Super simple config, only what we need need for now, we can expand, edit or separate as it grows.

I hard coded our specific defaults for ease, we eventually might want to inject them via env instead.


src/hooks/useChainId.js
```diff
@@ -0,0 +1,7 @@
+import env from '../environment'
+
+function useChainId() {
+  return env('CHAIN_ID')
+}
```
andy-hook, line 5: Here I was simply thinking ahead for run-time config. If we encourage using this hook it should be easier to update via state in the future.

src/lib/web3-utils.js
```diff
@@ -1,3 +1,12 @@
+export function getNetworkType(chainId) {
+  chainId = String(chainId)
+
+  if (chainId === '1') return 'main'
+  if (chainId === '4') return 'rinkeby'
```
facuspagnuolo, line 5: Is it possible to do something for staging? (cc @bpierre @sohkai)

bpierre: We could [pass a complete network](https://github.com/aragon/connect/blob/master/packages/connect-types/src/index.ts#L3-L7) to Connect in that case, rather than only the `chainId`, so we can change the `ensAddress`.

I’m also wondering if we should get rid of the name “networkType” completely, even for internal purposes? Instead we could only use `getNetworkConfig(chainId, variant)` where variant could be `undefined` or `"staging"`.

andy-hook: > I’m also wondering if we should get rid of the name “networkType” completely, even for internal purposes?.

I would like to but certain ui utils seem to still need it https://github.com/aragon/aragon-ui/blob/6d1b0f40f4029585b76ddfd7ac197c03c4fa303d/src/utils/web3.js#L13-L19

andy-hook: @facuspagnuolo  I added `rinkey-staging` and moved things around based on discussion with @bpierre, let me know what you think.

sohkai: @facuspagnuolo Just to double check, do you think we'll have multiple rinkeby orgs? E.g.

- Rinkeby <> Rinkeby court
- Staging (Rinkeby) <> Staging (Rinkeby) court

We would use the staging one primarily for development and leave the main Rinkeby org as a mirror of mainnet?

src/networks.js
```diff
@@ -0,0 +1,55 @@
...
+
+const DEFAULT_IPFS_GATEWAY = 'https://ipfs.eth.aragon.network/ipfs'
+const DEFAULT_ORG_LOCATION = 'agreement3.aragonid.eth'
+
+const NETWORK_ENVIRONMENTS = new Map([
+  [
+    'ethereum',
+    {
+      chainId: 1,
+      legacyNetworkType: 'main',
```
sohkai, line 11: Just wondering, when do we still need this (we should be eliminating it as much as possible 🔪)?

src/networks.js
```diff
@@ -0,0 +1,55 @@
...
+      ensAddress: '',
+      nodes: {
+        defaultEth: 'https://mainnet.eth.aragon.network/',
```
sohkai, line 14: I saw your comment in https://github.com/aragon/aragon/pull/1485 as to naming here, but I wonder if we should name this as "ethereum" if we are passing it down into the Connect configuration?

src/networks.js
```diff
@@ -0,0 +1,55 @@
+import env from './environment'
+
+const DEFAULT_IPFS_GATEWAY = 'https://ipfs.eth.aragon.network/ipfs'
+const DEFAULT_ORG_LOCATION = 'agreement3.aragonid.eth'
```
sohkai, line 4: I think we might as well specify these two settings to be per-network (especially the organization location).

src/networks.js
```diff
@@ -0,0 +1,55 @@
...
+    return JSON.parse(environment)
+  }
+
+  // Or use preset
+  return NETWORK_ENVIRONMENTS.get(environment)
+}
+
+export const orgLocation = env('ORG_LOCATION') || DEFAULT_ORG_LOCATION
+
+export const defaultIpfsGateway =
```
sohkai, line 54: After having to grapple with awkward dependency trees in the client, I wonder if we should keep this file as _pure_ configuration (without any imports or weird run-time side effects).

With the client, one of the changes I was going to do was move these configuration files into a `/configuration` folder, and maintain a `current-environment.js` that contains all the run-time selected defaults and settings.

src/networks.js
```diff
@@ -0,0 +1,55 @@
...
+const DEFAULT_IPFS_GATEWAY = 'https://ipfs.eth.aragon.network/ipfs'
+const DEFAULT_ORG_LOCATION = 'agreement3.aragonid.eth'
+
+const NETWORK_ENVIRONMENTS = new Map([
+  [
+    'ethereum',
+    {
+      chainId: 1,
+      legacyNetworkType: 'main',
+      ensAddress: '',
```
sohkai, line 12: No need to add this; Connect already has this handled for us!

src/networks.js
```diff
@@ -0,0 +1,55 @@
...
+        defaultEth: 'https://rinkeby.eth.aragon.network/',
+      },
+    },
+  ],
+])
+
+export function getNetworkEnvironment(environment) {
+  // If custom config supplied
+  if (environment.startsWith('{')) {
+    return JSON.parse(environment)
```
sohkai, line 45: Nitpick: we can try-catch and throw a better error message here.

src/networks.js
andy-hook: See https://github.com/aragon/network-dashboard/pull/8#discussion_r466980726

src/networks.js
andy-hook: Sure!

src/networks.js
andy-hook: Ah ok, I was told me might need to update it in certain cases!

src/networks.js
andy-hook: endpoints.ethereum? yes I totally prefer it, was trying to stick to our current standards and forgot about your PR 🙈

src/lib/web3-utils.js
facuspagnuolo: If possible I'd do that, we currently have different court instances so I think it would be interesting to support the same here. If not, let's start with staging for development now and then we can swap it to the Rinkeby one that mimics Mainnet

src/networks.js
andy-hook: I like this idea, in terms of reusability should we expect others who fork the project to update this configuration file in place of individual environment variables?

sohkai: > I like this idea, in terms of reusability should we expect others who fork the project to update this configuration file in place of individual environment variables

Oh yes, I think that would be a great way forward! Having a quick little guide like, "go here, edit this, set up your Vercel account, push" would be awesome once we want others to fork!

bpierre: What would you guys think about having all of these coming from the env vars, so that the entire app environment could be swapped without having to edit any source file? And we could use a default `.env` to have an equivalent of `current-environment.js`.

sohkai: I think this is already the case though, and we will still want some sort of `current-environment.js` file to call and export the results of `env()`?

bpierre: Oh I see, I thought we were talking about not using `process.env` anymore, but it’s more about having a convenient place to find all the environment-imported values. Sounds good :+1:

andy-hook: Ok, I made these changes, it's strictly build-time config now, I think it's fine and we shouldn't overthink it at this stage. Anything glaring let me know, else I'd like to get this in and move on :)

PR #10, opened 2020-08-17
Update font paths

andy-hook:
Fonts were showing decode errors and not rendering in sub pages, this was because of pathing.


src/hooks/disputable-voting-logic.js
```diff
@@ -3,16 +3,9 @@ import connectVoting from '@aragon/connect-voting-disputable'
...
 
   return votes
 }
-
-// Handles the main logic of the disputable voting app.
-export function useDisputableVotingLogic() {
-  return {
-    votes: useGetVotes(),
-  }
-}
```
andy-hook, line 18: I also streamlined this to keep us super simple for now.

PR #12, opened 2020-08-18
Add custom layout with breakpoints

andy-hook:
image

The designs require a larger width at the top end, this makes sense given we have more room without a sidebar. I've added initial breakpoints and the custom layout we should use to limit our views rather than the base provided by our UI kit.


src/breakpoints.js
```diff
@@ -0,0 +1,7 @@
+export const BREAKPOINTS = {
+  min: 360,
+  small: 680,
+  medium: 720,
+  large: 960,
+  max: 1280,
+}
```
andy-hook, line 7: How do we feel about the addition of this `max` breakpoint? I think it's needed given the size jump between medium and large with a higher value.

delfipolito: I like it!

src/components/Footer.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
-import { Layout, useTheme, GU } from '@aragon/ui'
+import { useTheme, GU } from '@aragon/ui'
+import Layout from './Layout'
 
 function Footer() {
```
delfipolito, line 5: I think we can remove this component for now cause Paty is still designing it

PR #13, opened 2020-08-18
Agreement: Add details

andy-hook:
image

Data to be populated from connect after the initial components are all available.


src/components/Agreement/Agreement.js
```diff
@@ -13,7 +15,14 @@ const Agreement = React.memo(function Agreement() {
       <Split
         primary={
           <>
-            <Box>Details</Box>
+            <Box>
+              <AgreementHeader title="Aragon Network DAO Agreement" />
+              <AgreementDetails
+                contractAddress="0x5c6620c49f9aecf74bd483054f2d0ace0d375f96"
+                creationDate="2020/07/20"
+                ipfsUri="ipfs:Qmb5CHbQQQx6YXkPE6HodeXVmtCRgpSgkj9EkW9xs6jDHj"
```
andy-hook, line 23: Pulled straight from old mock data, probably old and outdated but will be replaced.

src/components/Agreement/AgreementDetails.js
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+    </>
+  )
+}
+
+AgreementDetails.propTypes = {
+  ipfsUri: PropTypes.string,
+  creationDate: PropTypes.string,
+  contractAddress: PropTypes.string,
+}
```
andy-hook, line 69: I've stopped porting over propTypes and simply using base primitives to please the linter, this will all change in TS land anyway.

src/components/Agreement/Agreement.js
```diff
@@ -13,7 +15,14 @@ const Agreement = React.memo(function Agreement() {
       <Split
         primary={
           <>
-            <Box>Details</Box>
+            <Box>
+              <AgreementHeader title="Aragon Network DAO Agreement" />
+              <AgreementDetails
+                contractAddress="0x5c6620c49f9aecf74bd483054f2d0ace0d375f96"
+                creationDate="2020/07/20"
```
delfipolito, line 22: Maybe here we can pass the date in seconds like the one that is in the subgraph

src/components/Agreement/AgreementHeader.js
```diff
@@ -0,0 +1,83 @@
...
+          </h2>
+          <div
+            css={`
+              display: flex;
+            `}
+          >
+            <Tag
+              icon={<IconCheck size="small" />}
+              label="Active"
+              color={`${theme.positive}`}
```
delfipolito, line 45: Just wondering, can we just do `color={theme.positive}` here?

src/components/Agreement/InfoField.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
delfipolito, line 1: Just wondering, maybe we can move this component up so it can be used by both apps. What do you think? I have a very similar one on disputable voting app

src/components/Agreement/AgreementHeader.js
```diff
@@ -0,0 +1,83 @@
...
+            `}
+          >
+            {title}
+          </h2>
+          <div
+            css={`
+              display: flex;
+            `}
+          >
+            <Tag
```
delfipolito, line 42: Just checking, this tag should always appear? Or do we need a validation?

src/components/Agreement/AgreementHeader.js
andy-hook: Unfortunately `Tag` expects strings :(

src/components/Agreement/InfoField.js
andy-hook: Good idea! it's quite a common need.

src/components/Agreement/AgreementHeader.js
andy-hook: Good question, I think for the time being this can't enter any other state, there was discussion around whether we really need the tag at all but I do think it helps and is a nice piece of reassurance.

src/components/Agreement/Agreement.js
andy-hook: Definitely agree, how'd you feel if I do this in a separate improvement? I don't want to introduce too many more conflicts with date utils in our outstanding PRs.

src/components/Agreement/Agreement.js
andy-hook: Good idea! I think we're pretty close to getting the connector in so perhaps we can shape the data properly as we work on that :)

andy-hook:
Thanks for review @delfipolito, There are a couple of comments still to address but I will do those in separate improvements!

src/components/Agreement/Agreement.js
delfipolito: Sounds good!

PR #15, opened 2020-08-18
Add Agreement via static MD

andy-hook:
image


src/components/Agreement/AgreementDocument.js
```diff
@@ -0,0 +1,101 @@
...
+        <Markdown content={content} />
+      </Article>
+    </Box>
+  )
+}
+
+const Article = styled.article`
+  overflow-y: auto;
+  max-height: ${100 * GU}px;
+  padding: ${({ compact }) => `${compact ? 2 * GU : 7 * GU}px`};
```
andy-hook, line 23: Internal padding on the article to allow scrollbar to sit nicely against box edge

src/components/Agreement/static-agreement.js
```diff
@@ -0,0 +1,65 @@
+export default `
+# Aragon Network Agreement
```
andy-hook, line 2: I pulled this straight from the mock, it's most definitely incorrect or going to change.

PR #16, opened 2020-08-20
Agreement: Add static binding actions

andy-hook:
This PR adds the binding actions list as a static component (for now)

There is an issue with the expandable section height being incorrect, [this change](https://github.com/aragon/aragon-ui/pull/806) should remedy it.

There are no badge graphics nor app addresses being passed to them, this is best handled when supplying data via connect.

@delfipolito I also addressed your comment about temporarily centralising into a mock data file for the agreement.

image


src/components/InfoField.js
```diff
@@ -1,23 +1,27 @@
...
-      </h3>
-      {children}
-    </div>
+    <Field
+      label={label}
+      {...props}
+      css={`
+        margin-bottom: 0;
+      `}
+    >
```
andy-hook, line 13: This change is to inherit spacing certain spacing properties such as when adding a `<Help/>` element to a label as the UI kit already has opinions about that.

src/components/Agreement/mock-data.js
```diff
@@ -0,0 +1,20 @@
...
+  collateralToken: {
+    address: '0x6b175474e89094c44da98b954eedeac495271d0f',
+    decimals: 18,
+    name: 'DAI Token',
+    symbol: 'DAI',
+  },
+  actionAmount: '100000000000000000000',
+  challengeAmount: '200000000000000000000',
+  challengeDuration: 172800000,
+}
```
andy-hook, line 13: This is based on the outdated schema for the client, I wanted to port this over quickly and it'll be easier to update when we do the connect work.

src/components/Agreement/AgreementBindingActions.js
```diff
@@ -0,0 +1,177 @@
...
+
+const AgreementBindingActions = React.memo(({ apps }) => {
+  const items = apps.map(
+    ({ actionCollateral, appName, challengeCollateral, settlementPeriod }) => [
+      <div
+        key="badge"
+        css={`
+          display: flex;
+          align-items: center;
+          margin-left: ${-1 * GU}px;
```
andy-hook, line 23: Visual adjustment to make this site nicer against the expansion button (I think it looks tighter)

PR #17, opened 2020-08-21
Proposals detail

delfipolito:
image


src/components/Proposals/ProposalCard.js
```diff
@@ -1,23 +1,83 @@
...
+          // the latest versions of Safari (as of June 2020), which
+          // is why word-break:break-word has been added here.
+          hyphens: auto;
+          overflow-wrap: anywhere;
+          word-break: break-word;
+        `}
+      >
+        <span css="font-weight: bold">#{voteId}: </span>
+        <span>{context || 'No description provided'}</span>
+      </div>
```
andy-hook, line 35: Should we make this a paragraph with `<strong>` for bolding? I also think we have room to make this description larger, in the designs it looks like 18px or `textStyle("body1")`

src/components/Proposals/ProposalCard.js
```diff
@@ -1,23 +1,83 @@
...
+        }
+      />
+      <div
+        css={`
+          margin-top: ${2 * GU}px;
+        `}
+      >
+        <DisputableStatusLabel
+          status={DISPUTABLE_VOTE_STATUSES.get(vote.status)}
+        />
```
andy-hook, line 61: I was wondering about the style of "Rejected" and "Passed" and whether it should be using the same visual as the pill labels. Looking at the designs the style is actually different for these two.

image

Depending on the use cases for the `DisputableStatusLabel` it might be worth separating "Rejected" and "Passed" from that component and updating the style to match.

src/components/Proposals/ProposalCardGroup.js
```diff
@@ -0,0 +1,61 @@
...
+  const rowHeight = compactMode ? null : 294
+
+  return (
+    <section>
+      <h2
+        css={`
+          display: flex;
+          align-items: center;
+          margin-bottom: ${3 * GU}px;
+          ${compactMode ? `padding: 0 ${2 * GU}px;` : ''}
```
andy-hook, line 26: Tiny nitpick but we could the `&&` operator to avoid the empty string.

src/components/Proposals/ProposalOption.js
```diff
@@ -0,0 +1,70 @@
...
+          css={`
+            ${textStyle('label2')}
+          `}
+        >
+          {label}
+        </span>
+        <span
+          css={`
+            ${textStyle('body3')}
+            font-size: 12px;
```
andy-hook, line 21: We can just use `textStyle("body4")` for this size.

src/components/Proposals/ProposalOption.js
```diff
@@ -0,0 +1,70 @@
...
+          border-radius: ${RADIUS}px;
+        `}
+      >
+        <Bar
+          css={`
+            background-color: ${color || theme.positive};
+            width: 100%;
+            transform: scale3d(${percentage / 100}, 1, 1);
+          `}
+        />
```
andy-hook, line 41: Would [ProgressBar](https://ui.aragon.org/progress-bar/) work here?

src/components/Proposals/ProposalCard.js
```diff
@@ -1,23 +1,83 @@
...
+        }
+      />
+      <ProposalOption
+        color={theme.negative}
+        percentage={(vote.nays * 100) / vote.votingPower}
+        label={
+          <WrapProposalOption>
+            <span>No</span>
+          </WrapProposalOption>
+        }
```
andy-hook, line 52: Do we anticipate needing to supply other more complex labels? If not it might be easier to nest `WrapProposalOption` styles within the component and pass less in the prop.

andy-hook (approved):
Just a few design related comments :) are we going to add `AppBadge` here or in a separate improvement?

src/components/Proposals/ProposalCard.js
delfipolito: Yes :) I left the `css="font-wight: bold` tough, cause otherwise aragonUI overrides the strong and b tag with font-weight: 600

src/components/Proposals/ProposalCardGroup.js
bpierre: Just a note on this: using `&&` would be interpreted as `"false"` when `compactMode` is `false`, which can create issues with the way the CSS block gets interpreted − especially since the `;` would be missing here. That’s why we always use the ternary operator rather than `&&` (except if the condition itself is a string of course) :smile:

andy-hook: Oh that's interesting, I didn't realise that was the case! I assumed styled components was smart enough to strip it out! shall commit that to memory :)

bpierre: Yes I was surprised too the first time. I think they should remove values like `null`, `true`, `false`, similar to how JSX does it. Actually I haven’t checked in a while, so it might be worth checking that it’s still the case!

src/components/Proposals/ProposalCard.js
andy-hook: Is font-weight: 600 not bold enough? Perhaps we should consider a global style for `strong` inside a paragraph? I can see us needing to use it a lot.

src/components/Proposals/DisputableStatusLabel.js
```diff
@@ -72,10 +74,36 @@ function DisputableStatusLabel({ status }) {
...
+          ${textStyle('body2')};
+          color: ${color || theme.surfaceContentSecondary};
+          display: flex;
+          align-items: center;
+        `}
+      >
+        {Icon && <Icon size="small" />}
+        <span
+          css={`
+            margin-left: $ ${0.5 * GU}px;
```
andy-hook, line 94: Hey @delfipolito I think this is a typo.

PR #9, opened 2020-08-12
[WIP] Proposal detail screen 

delfipolito:
image

PR #22, opened 2020-08-24
Proposal detail: Backgrounds based on status

delfipolito:
image

image

image

### DIFF
--- a/src/components/Proposals/FeedbackModule.js
+++ b/src/components/Proposals/FeedbackModule.js
@@ -9,26 +9,11 @@ import {
   textStyle,
   useTheme,
 } from '@aragon/ui'
-import { addressesEqual } from '../../lib/web3-utils'
 import { dateFormat, toMs } from '../../lib/date-utils'
 
-function FeedbackModule({ vote, connectedAccount }) {
+function FeedbackModule({ vote, mode }) {
   // TODO: Replace dates and amounts with real data
   const theme = useTheme()
-
-  let mode = null
-
-  if (vote.challenger && addressesEqual(vote.challenger, connectedAccount)) {
-    mode = 'challenger'
-  }
-
-  if (addressesEqual(vote.creator, connectedAccount)) {
-    mode = 'submitter'
-  }
-
-  if (!mode) {
-    return <div />
-  }
 
   return (
     <div
@@ -114,7 +99,7 @@ function Strong({ children }) {
 
 FeedbackModule.propTypes = {
   vote: PropTypes.object.isRequired,
-  connectedAccount: PropTypes.string,
+  mode: PropTypes.string,
 }
 
 Strong.propTypes = {
--- a/src/components/Proposals/FeedbackModule.js
+++ b/src/components/Proposals/FeedbackModule.js
@@ -4,6 +4,7 @@ import {
   GU,
   IconAttention,
   IconCoin,
+  Link,
   RADIUS,
   textStyle,
   useTheme,
@@ -17,10 +18,7 @@ function FeedbackModule({ vote, connectedAccount }) {
 
   let mode = null
 
-  if (
-    vote.challenge &&
-    addressesEqual(vote.challenge.challenger, connectedAccount)
-  ) {
+  if (vote.challenger && addressesEqual(vote.challenger, connectedAccount)) {
     mode = 'challenger'
   }
 
@@ -38,7 +36,6 @@ function FeedbackModule({ vote, connectedAccount }) {
         border-radius: ${RADIUS}px;
         background: ${theme.background};
         padding: ${3.5 * GU}px ${10 * GU}px;
-        text-align: center;
       `}
     >
       <div
@@ -47,7 +44,6 @@ function FeedbackModule({ vote, connectedAccount }) {
           grid-template-columns: auto 1fr;
           grid-gap: ${3 * GU}px;
           align-items: center;
-          text-align: left;
         `}
       >
         <div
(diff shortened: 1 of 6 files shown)


src/components/Proposals/FeedbackModule.js
```diff
@@ -17,10 +18,7 @@ function FeedbackModule({ vote, connectedAccount }) {
 
   let mode = null
 
-  if (
-    vote.challenge &&
-    addressesEqual(vote.challenge.challenger, connectedAccount)
-  ) {
+  if (vote.challenger && addressesEqual(vote.challenger, connectedAccount)) {
```
andy-hook, line 21: Should we raise this logic up as described [here](https://github.com/aragon/network-dashboard/pull/9/files#r474687155)?

src/components/Proposals/FeedbackModule.js
```diff
@@ -101,18 +98,18 @@ function FeedbackModule({ vote, connectedAccount }) {
   )
 }
 
-function Bold({ children }) {
+function Strong({ children }) {
```
andy-hook, line 101: I'm not sure we need this component, I think the globals are already providing these styles for paragraph and strong tags.

src/components/Proposals/ProposalDetail.js
```diff
@@ -17,12 +17,47 @@ import {
...
 import SummaryBar from './SummaryBar'
+import FeedbackModule from './FeedbackModule'
 import Layout from '../Layout'
 import { networkEnvironment } from '../../current-environment'
 
+function getAttributes(status, theme) {
+  const attributes = {
+    [VOTE_STATUS_CANCELLED]: {
+      background: '#F9FAFC',
+      border: '#DDE4E9',
```
andy-hook, line 37: If these colours are not available in the base theme perhaps we should extend our own theme object to avoid repeating them in multiple places throughout the app?

src/components/Proposals/SummaryBar.js
```diff
@@ -4,23 +4,34 @@ import { RADIUS, useTheme } from '@aragon/ui'
...
   return (
     <Main {...props}>
       <CombinedBar>
         {!!positiveSize && (
           <BarPart
             style={{
-              backgroundColor: theme.positive,
+              backgroundColor: `${
+                disabled ? theme.surfaceOpened : theme.positive
+              }`,
```
andy-hook, line 24: For consistency perhaps we should use the `css` prop to apply these, rather than `style`.

src/components/Proposals/InfoBoxes.js
```diff
@@ -14,7 +14,7 @@ import { dateFormat, toMs } from '../../lib/date-utils'
 import { round, safeDiv } from '../../lib/math-utils'
 import SummaryBar from './SummaryBar'
 
-function InfoBoxes({ vote }) {
+function InfoBoxes({ vote, disabled }) {
```
andy-hook, line 17: I don't think this is a problem yet, but if we see ourselves drilling lots of props down to these components it might be worth flattening it out a providing each component with their props directly from within `ProposalDetail`

src/components/Proposals/ProposalDetail.js
```diff
@@ -17,12 +17,47 @@ import {
...
+      disabled: true,
+    },
+    [VOTE_STATUS_PAUSED]: {
+      background: '#fffdfa',
+      border: '#F5A623',
+      disabled: true,
```
andy-hook, line 43: Do you think we could be a bit more descriptive with these property names. How about `backgroundColor`, `borderColor`, `disabledProgressBars` ?

andy-hook (commented):
Left a few comments, nothing major. We should start to think about how we're managing colours and common styles, we're likely going to need to use these values in other places so should aim for consistency :)

src/components/Proposals/ProposalDetail.js
delfipolito: good idea!

src/components/Proposals/FeedbackModule.js
delfipolito: We need it because we also need to change font color

andy-hook: ah, I thought paragraphs were already defaulting to `surfaceContent`

src/components/Proposals/ProposalDetail.js
andy-hook: I think it would be good to update our existing palette with the new values if they are different in the design. If you look at the [colours page on devbox](https://devbox.dweb.cool/#Colors) we have a lot of crossover between the various `warning` `positive` `negative`, we can update them with new values in our [theme](https://github.com/aragon/network-dashboard/blob/master/src/theme.js) and access them with `useTheme`.

For example the `challenged` colours are very close to `warning`, we could just assign them to `warning` in our theme and use them from there, rather than having custom hex values in the templates themselves.

src/components/Proposals/ProposalDetail.js
```diff
@@ -17,11 +17,47 @@ import {
...
+      backgroundColor: theme.surfacePressed,
+      borderColor: theme.controlUnder,
+      disabledProgressBars: true,
+    },
+    [VOTE_STATUS_PAUSED]: {
+      backgroundColor: '#fffdfa',
+      borderColor: theme.warning,
+      disabledProgressBars: true,
+    },
+    [VOTE_STATUS_DISPUTED]: {
```
andy-hook, line 46: I was just looking through the designs and wondering about the term 'disputed' now, it seems like some of the phrasing has changed to 'challenged'? do you think we should we update our naming to reflect this?

andy-hook (approved):

PR #24, opened 2020-08-25
Background by state & app badges on proposal cards

delfipolito:
image


src/components/Proposals/ProposalCard.js
```diff
@@ -1,25 +1,75 @@
...
         grid-gap: ${1 * GU}px;
         padding: ${3 * GU}px;
+        background: ${backgroundColor};
+        border: solid 1px ${borderColor};
       `}
     >
+      <div
+        css={`
+          display: flex;
+          justify-content: space-between;
```
andy-hook, line 67: I have a feeling we don't actually need these flex rules as it should just sit at the start of the container without them. If not we can also use `justify-content: flex-start` to position it at the very beginning (which I think is the default for flex anyway).

src/components/Proposals/ProposalCard.js
```diff
@@ -1,25 +1,75 @@
...
+          margin-bottom: ${1 * GU}px;
+        `}
+      >
+        <AppBadge label="Disputable voting" appAddress={appAddress} />
```
andy-hook, line 71: Nice! i'm working on getting this information easily at the moment.

src/components/Proposals/ProposalCard.js
```diff
@@ -1,25 +1,75 @@
...
+      </div>
       <p
         css={`
           // overflow-wrap:anywhere and hyphens:auto are not supported yet by
```
andy-hook, line 75: Would we be able to align the paragraph vertically at the start of the container so that it always sits close under the badge? that way all of the texts should align nicely along the row as well.

src/hooks/disputable-voting-logic.js
```diff
@@ -8,7 +8,7 @@ export function useVotes() {
   const [disputableVoting] = useApp('disputable-voting')
   const [votes] = useDisputableVoting(disputableVoting, (app) => app.votes())
 
-  return votes
+  return { app: disputableVoting, votes: votes }
```
andy-hook, line 11: Should we export / import the app as `disputableVotingApp` just to be explicit? :)

andy-hook (approved):
Couple of minor things but LGTM!

PR #25, opened 2020-08-26
Initial agreement details via connector

andy-hook:
This PR adds some initial data for the Agreement and also reorganises how we are accessing data for the disputable votes.

These are initial changes ahead of a rework to support the new React api, as such they work around issues we had with the previous version, this will improve with upcoming changes.

I've front-loaded `votes` and `agreement` by storing them in context, this avoids us having to show loading messages every time a route changes. This is not perfect and not a best practice, in the future I recommend we progressively render these views and request on demand, but for now this will get us to the initial version faster and should work fine.

@delfipolito I still have more agreement data to pull, I will also add a way to access icons and formatted app names soon. I slimmed down error and loading handling because they will be reliably provided by react api thus we won't have to manage as much on our side. I will merge this straight away to continue as fast as possible, hope this is ok.


src/App.js
```diff
@@ -1,20 +1,26 @@
...
     <Connect>
-      <ViewportProvider>
-        <Router>
-          <MainView>
-            <Routes />
-          </MainView>
-        </Router>
-      </ViewportProvider>
+      <OrgAppsProvider>
+        <AppDataProvider>
```
andy-hook, line 14: In the future we will remove `AppDataProvider` and move the requests into smaller to progressively request things in views as we need. I still think it is a nice idea to provide the base apps via `OrgAppsProvider` though.

src/components/MainView.js
```diff
@@ -2,8 +2,31 @@ import React from 'react'
...
+  if (loading) {
+    return (
+      <div
+        css={`
+          display: flex;
+          align-items: center;
+          justify-content: center;
+          height: 100vh;
+          width: 100%;
+        `}
```
andy-hook, line 23: This will soon be updated to a nice looking loader from the designs.

src/configuration/networks.js
```diff
@@ -21,7 +20,6 @@ export default new Map([
         ethereum: 'https://rinkeby.eth.aragon.network/',
       },
       orgLocation: 'agreement3.aragonid.eth',
-      disputableVotingApp: '0x26e14ed789b51b5b226d69a5d40f72dc2d0180fe',
```
andy-hook, line 24: I removed this for now as we are waiting for the apps list before rendering the app and can get it dynamically.

src/providers/OrgApps.js
```diff
@@ -0,0 +1,50 @@
...
+function getAppByName(apps, appName) {
+  return apps.find(({ name }) => name === appName) || null
+}
+
+function OrgAppsProvider({ children }) {
+  const [apps] = useApps()
+
+  // Avoid additional overhead by finding within existing app list
+  const agreementApp = getAppByName(apps, 'agreement')
+  const disputableVotingApp = getAppByName(apps, 'disputable-voting')
```
andy-hook, line 21: It felt unnecessary to use `useApp` multiple times when we have to get the full list anyway.

PR #26, opened 2020-08-26
Environment: Upgrade to latest beta connectors

andy-hook:
@delfipolito This change should hopefully bring an end to local linking and branch management on connectors :)

Edit: Remember to unlink any existing syms, it's probably best to nuke `node_modules` as well

### DIFF
--- a/now.json
+++ b/now.json
@@ -7,7 +7,7 @@
   },
   "build": {
     "env": {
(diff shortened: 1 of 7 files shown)


andy-hook:
CI finally passing again, going to merge and continue, will address any comments in separate improvements.

PR #30, opened 2020-08-26
Agreement: Extend disputableApps with collateral requirements

andy-hook:
Extends `disputableApps` to display complete Binding Action details via connector.

### DIFF
--- a/src/hooks/useAgreementDetails.js
+++ b/src/hooks/useAgreementDetails.js
@@ -1,8 +1,10 @@
 import { useEffect, useState } from 'react'
-import { useOrgApps } from '../providers/OrgApps'
 import { createAppHook } from '@aragon/connect-react'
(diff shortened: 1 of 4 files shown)


src/hooks/useAgreementDetails.js
```diff
@@ -81,7 +70,58 @@ export function useAgreementDetails() {
...
+}
+
+async function getExtendedDisputableApps(apps, disputableApps) {
+  const allRequirements = await Promise.all(
+    disputableApps.map((app) => app.collateralRequirement())
+  )
+
+  const allTokens = await Promise.all(
+    allRequirements.map((collateral) => collateral.token())
+  )
```
andy-hook, line 97: Perhaps this could be made easier (and faster) at the subgraph level.

PR #44, opened 2020-08-27
Decode script to get description

delfipolito:
image

image

### DIFF
--- a/src/hooks/useDisputableVotes.js
+++ b/src/hooks/useDisputableVotes.js
@@ -1,9 +1,11 @@
 import { useEffect, useState } from 'react'
+import { describeScript } from '@aragon/connect'
 import connectVoting from '@aragon/connect-disputable-voting'
-import { createAppHook } from '@aragon/connect-react'
+import { useApps, createAppHook } from '@aragon/connect-react'
 import { useOrgApps } from '../providers/OrgApps'
 import { networkEnvironment } from '../current-environment'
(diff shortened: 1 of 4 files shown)


src/hooks/useDisputableVotes.js
```diff
@@ -15,20 +17,84 @@ const useDisputableVotingHook = createAppHook(connectVoting, connecterConfig)
...
+          setVotes(processedVotes)
+          setLoading(false)
+        }
+      } catch (error) {
+        console.error(error)
+      }
+    }
+
+    if (!votesLoading) {
+      getExtendedVotes()
```
andy-hook, line 55: `getProcessedVotes` ?

andy-hook (changes requested):
This is looking really great, and is going to add a lot to our vote presentation 😍 

It would be good to figure out what was happening with the dependencies in `useEffect`, I think the switch to pulling `apps` from the existing context will help with that.

One thing I totally forgot to mention and that might have added confusion is the way that parcel handles hot module reloading with react context, you might have seen `undefined / null` errors after saving which disappeared with a fresh reload, it's fiddly and awkward and only happens with these data providers. I can explain more offline.

Let me know what you think!

andy-hook (approved):
I've addressed my comments with the exception of one regarding naming but not a big deal! hope this is ok, will merge to keep us going :)

PR #47, opened 2020-08-28
Proposal detail: Show both justification and description

delfipolito:
image


src/hooks/useDisputableVotes.js
```diff
@@ -143,6 +143,6 @@ async function processVote(vote, apps) {
 
   return {
     ...extendedVote,
-    metadata: description,
+    description: description,
```
andy-hook, line 146:
```suggestion
    description,
```

If you prefer :)

andy-hook (approved):

PR #52, opened 2020-08-31
Banner: Change copy for PC

delfipolito:
image

### DIFF
--- a/src/components/Proposals/ProposalBanner.js
+++ b/src/components/Proposals/ProposalBanner.js
@@ -62,7 +62,7 @@ function ProposalBanner({ onCloseBanner, ...props }) {
             margin-bottom: ${1.5 * GU}px;
           `}
         >
-          Fight For Freedom
+          This is a test
         </h1>
         <div
           css={`
@@ -73,10 +73,9 @@ function ProposalBanner({ onCloseBanner, ...props }) {
             color: ${theme.contentSecondary};
           `}
         >
-          This is a test <br /> Aragon Network Cash is a test organization set
-          up as part of a new round of precedence campaign disputes. Learn more
-          on the <Link href="https://aragon.org/blog">Aragon project blog</Link>
-          .
+          Aragon Network Cash is a test organization set up as part of a new
+          round of precedence campaign disputes. Learn more on the{' '}
+          <Link href="https://aragon.org/blog">Aragon project blog</Link>.
         </div>
         <div
           css={`
--- a/src/components/Proposals/ProposalBanner.js
+++ b/src/components/Proposals/ProposalBanner.js
@@ -5,6 +5,7 @@ import {
   ButtonIcon,
   GU,
   IconCross,
+  Link,
   textStyle,
   useTheme,
 } from '@aragon/ui'
@@ -63,7 +64,7 @@ function ProposalBanner({ onCloseBanner, ...props }) {
         >
           Fight For Freedom
         </h1>
-        <p
+        <div
           css={`
             ${textStyle('body1')};
             max-width: ${95 * GU}px;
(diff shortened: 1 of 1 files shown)


src/components/Proposals/ProposalBanner.js
```diff
@@ -73,19 +72,40 @@ function ProposalBanner({ onCloseBanner, ...props }) {
             color: ${theme.contentSecondary};
           `}
         >
-          We believe humankind should use technology as a liberating tool to
-          unleash all the goodwill and creativity of our species, rather than as
-          a tool to enslave and take advantage of one another.
+          This is a test <br /> Aragon Network Cash is a test organization set
```
andy-hook, line 75: Is this the final campaign copy? it reads a little strange to me in this configuration. Should "This is a test" be in the paragraph? or is it intended as the title?

src/components/Proposals/ProposalBanner.js
```diff
@@ -73,19 +72,40 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+          up as part of a new round of precedence campaign disputes. Learn more
+          on the Aragon project blog
```
andy-hook, line 77:
```suggestion
          up as part of a new round of precedence campaign disputes. Learn more
          on the Aragon project blog.
```

I'm also wondering should we link to the post in this same sentence? e.g:

Learn more on the [Aragon project blog](https://aragon.org/blog).

src/components/Proposals/ProposalBanner.js
```diff
@@ -73,19 +72,40 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+            display: flex;
+            flex-wrap: wrap;
+            align-items: center;
+            justify-content: center;
           `}
-        />
+        >
+          <Button
+            label="Learn more"
+            mode="normal"
```
andy-hook, line 89: As this is the first action in the pair (and more important) it should probably be `mode="strong"` with the manifesto link becoming the secondary.

src/components/Proposals/ProposalBanner.js
```diff
@@ -73,19 +72,40 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+            href="https://aragon.org/blog"
```
andy-hook, line 90: Is there a specific blog post or area we can link to about this campaign?

src/components/Proposals/ProposalBanner.js
```diff
@@ -73,19 +72,40 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+            css={`
+              padding: ${2.75 * GU}px ${3.5 * GU}px;
+              font-weight: 600;
+              margin: ${3 * GU}px ${2 * GU}px 0;
```
andy-hook, line 94: We should move the `margin-top` to the parent and internally space the buttons separately. I also feel the gap between the buttons is a little too generous and could be tighter, what do you think?

andy-hook (commented):
Thanks for adding this! I was wondering about the content and whether we have it quite right. "This is a test" looks a bit strange to me like this. @john-light is this all accurate?

src/components/Proposals/ProposalBanner.js
delfipolito: Yes, but the post is not out yet, so im gonna link it in the last minute to the specific post

src/components/Proposals/ProposalBanner.js
delfipolito: Margin top is there to separate the  buttons from each others on mobile

andy-hook (approved):
Tweaked the button spacing a touch, otherwise LGTM, let's check in with @john-light about the blog link, copy question and amend later if needed.

src/components/Proposals/ProposalBanner.js
andy-hook: Sounds good, we can adjust this after we have better responsive utils merged.

PR #54, opened 2020-08-31
Proposal cards: App badges

delfipolito:
image

### DIFF
--- a/src/components/Proposals/ProposalCard.js
+++ b/src/components/Proposals/ProposalCard.js
@@ -41,7 +41,7 @@ function getAttributes(status, theme) {
 
 function ProposalCard({ appAddress, vote, onProposalClick }) {
   const theme = useTheme()
-  const { context, voteId, description } = vote
+  const { context, voteId, description, target } = vote
 
   const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
   const { backgroundColor, borderColor, disabledProgressBars } = getAttributes(
@@ -69,7 +69,14 @@ function ProposalCard({ appAddress, vote, onProposalClick }) {
           margin-bottom: ${1 * GU}px;
         `}
       >
-        <AppBadge label="Disputable Voting" appAddress={appAddress} badgeOnly />
+        {target && (
+          <AppBadge
+            label={target.name ? target.name : target.address}
+            appAddress={target.address}
+            iconSrc={target.icon ? target.icon : ''}
+            badgeOnly
+          />
+        )}
       </div>
 
       <p
--- a/src/components/Proposals/ProposalCard.js
+++ b/src/components/Proposals/ProposalCard.js
@@ -69,14 +69,12 @@ function ProposalCard({ appAddress, vote, onProposalClick }) {
           margin-bottom: ${1 * GU}px;
         `}
       >
-        {target && (
-          <AppBadge
-            label={target.name ? target.name : target.address}
-            appAddress={target.address}
-            iconSrc={target.icon ? target.icon : ''}
-            badgeOnly
-          />
-        )}
+        <AppBadge
+          label={target.name ? target.name : target.address}
+          appAddress={target.address}
+          iconSrc={target.icon ? target.icon : ''}
+          badgeOnly
+        />
       </div>
 
       <p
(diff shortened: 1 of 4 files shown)


src/hooks/useDisputableVotes.js
```diff
@@ -133,16 +136,40 @@ async function processVote(vote, apps) {
...
     status: vote.status,
+    target: getTargetData(apps, disputableVotingApp.address),
   }
 
   if (vote.script === EMPTY_SCRIPT) {
     return extendedVote
   }
 
   const description = await describeScript(vote.script, apps)
+  if (description[0] && description[0].to) {
```
andy-hook, line 147: Should we destructure / assign this on a previous line to save us accessing via the index each time?

src/hooks/useDisputableVotes.js
```diff
@@ -133,16 +136,40 @@ async function processVote(vote, apps) {
...
+      icon: '',
+    }
+  }
+
+  if (
+    apps &&
+    apps.filter((app) => app.address === extendedVote.target.address).length > 0
+  ) {
+    extendedVote.target = getTargetData(apps, extendedVote.target.address)
+  }
```
andy-hook, line 160: Find this logic a little bit confusing, perhaps we could clarify the intention by assigning these as descriptive variables / adding comments? Eventually we might want to break them up into composable functions but in the interest of time just a little bit more description here would be useful.

andy-hook (commented):
Can't wait to get this in the app 😍

I've added a few small things, the logic of getting the appearance is a little confusing by the nature of the checks. I think making it more descriptive would be good in the short term so we have it ready for the PC, but we will probably want to re-think how it's structured as to make it clearer for the future.

src/hooks/useDisputableVotes.js
delfipolito: i tried to explain this a little bit more. let me know if this is ok :)

src/components/Proposals/ProposalCard.js
```diff
@@ -69,7 +69,14 @@ function ProposalCard({ appAddress, vote, onProposalClick }) {
...
       >
-        <AppBadge label="Disputable Voting" appAddress={appAddress} badgeOnly />
+        {target && (
+          <AppBadge
+            label={target.name ? target.name : target.address}
+            appAddress={target.address}
+            iconSrc={target.icon ? target.icon : ''}
+            badgeOnly
+          />
+        )}
```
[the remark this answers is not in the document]
delfipolito: is not necessary anymore

src/hooks/useDisputableVotes.js
```diff
@@ -133,16 +136,46 @@ async function processVote(vote, apps) {
     naysPct: vote.naysPct,
     yeasPct: vote.yeasPct,
     status: vote.status,
+    target: getTargetData(apps, disputableVotingApp.address),
   }
 
+  // return with disputableVotingApp as the default app target
```
andy-hook, line 142:
```suggestion
  // Return disputableVotingApp as the default app target
```

src/hooks/useDisputableVotes.js
```diff
@@ -133,16 +136,46 @@ async function processVote(vote, apps) {
...
   if (vote.script === EMPTY_SCRIPT) {
     return extendedVote
   }
 
   const description = await describeScript(vote.script, apps)
+  const voteHasAppTarget = description[0] && description[0].to
+
+  // if we have the target app on the description, we change the default one
```
andy-hook, line 150:
```suggestion
  // If we have the target app on the description, build the target data with it instead
```

src/hooks/useDisputableVotes.js
```diff
@@ -133,16 +136,46 @@ async function processVote(vote, apps) {
...
+  if (voteHasAppTarget) {
+    extendedVote.target = {
+      address: description[0].to,
+      name: description[0].name || description[0].identifier,
+    }
+  }
+
+  // if the target app is one of the known ones
+  // we try to get the icon and name from there
```
andy-hook, line 159:
```suggestion
  // If the target app is known to us then we can get it's icon and name 
```

andy-hook (approved):
Couple of amends to the comments, are these accurate?

PR #56, opened 2020-08-31
Disputed status style

delfipolito:
- Changed the colors of the backgrund and strokes for the disputable status
- Added a new custom icon for disputable status

image


src/assets/disputed.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="13" height="13" viewBox="0 0 13 13" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect x="4.82837" y="5.70697" width="5.04901" height="4.06286" transform="rotate(-45.5209 4.82837 5.70697)" fill="#FF7C7C"/>
+<rect x="4.63574" y="7.53192" width="1.75352" height="5.21479" transform="rotate(44.341 4.63574 7.53192)" fill="#FF7C7C"/>
+<path d="M3.46909 4.476C3.26651 4.27707 3.26355 3.9516 3.46248 3.74902L6.2797 0.880102C6.47862 0.677529 6.8041 0.674569 7.00667 0.873492L7.63345 1.48898L4.09587 5.09148L3.46909 4.476Z" fill="#FF7C7C"/>
+<path d="M8.45996 9.22961L11.9975 5.62711L12.6243 6.2426C12.8269 6.44152 12.8299 6.767 12.6309 6.96957L9.81372 9.83849C9.61479 10.0411 9.28932 10.044 9.08674 9.8451L8.45996 9.22961Z" fill="#FF7C7C"/>
+</svg>
```
andy-hook, line 6: We will probably want to standardise and add this to the ui kit but for now shall we mimic the [aragon/ui implementation](https://github.com/aragon/aragon-ui/blob/master/src/icons/components/IconArrowUp.js)?

We will want to optimise the SVG code, make sure the `viewBox` and `currentColor` usage matches and place it in `./icons`

src/components/Proposals/DisputableStatusLabel.js
```diff
@@ -22,47 +21,48 @@ import {
...
       label: 'Passed',
-      Icon: IconCheck,
+      icon: <IconCheck size="small" />,
       color: theme.positive,
     },
     [VOTE_STATUS_CANCELLED]: {
       background: theme.surfaceUnder,
       label: 'Cancelled',
-      Icon: IconClose,
+      icon: <IconClose size="small" />,
```
andy-hook, line 41: If we make the above change to the icon we won't have to update this code, it should just slot into the previous implementation.

andy-hook (commented):
Nice! only comments are about the icon, if we make it a component that matches the standard format provided by the ui kit then it can slot in without any extra effort 👍  then we can raise a PR in that project to have it added to the main set.

src/assets/disputed.svg
delfipolito: good idea!

andy-hook (approved):

PR #58, opened 2020-08-31
Ipfs link on justifications

delfipolito:
image

### DIFF
--- a/src/components/Proposals/ProposalDetails/ProposalDetails.js
+++ b/src/components/Proposals/ProposalDetails/ProposalDetails.js
@@ -31,7 +31,7 @@ import StatusInfo from './StatusInfo'
 import FeedbackModule from './FeedbackModule'
 import Description from '../Description'
 import { addressesEqual } from '../../../lib/web3-utils'
-import { getIpfsCidFromUri, getIpfsUrlFromUri } from '../../../lib/ipfs-utils'
+import { getIpfsUrlFromUri } from '../../../lib/ipfs-utils'
 
 function getAttributes(status, theme) {
   const attributes = {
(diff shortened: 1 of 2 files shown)


src/components/Proposals/ProposalDetails/ProposalDetails.js
```diff
@@ -152,7 +158,27 @@ function Details({ vote, status }) {
...
+              >
+                <span
+                  css={`
+                    display: block;
+                    overflow: hidden;
+                    text-overflow: ellipsis;
+                    text-align: left;
+                  `}
+                >
+                  {getIpfsCidFromUri(ipfsJustification)}
```
Evalir, line 176: Should we show "Read more" instead of the IPFS URI?

delfipolito: Done :)

src/lib/web3-utils.js
```diff
@@ -25,9 +25,11 @@ export function getAppPresentation(apps, appAddress) {
...
     ({ address }) => address === appAddress
   )
-
-  const iconPath = manifest.icons[0].src
-  const iconSrc = getIpfsUrlFromUri(contentUri) + iconPath
+  let iconSrc = ''
+  if (manifest.icons && manifest.icons[0]) {
+    const iconPath = manifest.icons[0].src
+    iconSrc = getIpfsUrlFromUri(contentUri) + iconPath
+  }
```
andy-hook, line 32: Is this change supposed to be here? I can't see how it applies.

delfipolito: Yes, sorry i put it here cause i change the org to the one that we will be  using and some of them doesnt have manifest.icons so it was breaking

andy-hook (approved):

PR #61, opened 2020-08-31
Capture breaking errors via Sentry

andy-hook:
Will build on this with better UX for handling errors in separate improvements.


sohkai (approved):
🙌

PR #67, opened 2020-09-01
Precedence campaign banner

delfipolito:
image


src/components/Proposals/ProposalBanner.js
```diff
@@ -39,43 +42,62 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+                line-height: 1.1;
+                font-weight: 700;
+                text-align: left;
+                margin-bottom: ${2.5 * GU}px;
+                max-width: ${62 * GU}px;
+              `}
+            >
+              This is a test organization
+            </h1>
+            <div
```
andy-hook, line 72: Should we make this a paragraph?

src/components/Proposals/ProposalBanner.js
```diff
@@ -39,43 +42,62 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+              css={`
+                ${textStyle('title4')};
+                font-size: ${compactMode ? 2.5 * GU : 2.75 * GU}px;
+                max-width: ${62 * GU}px;
+                text-align: left;
```
andy-hook, line 77: If we remove the text-align from the container then we won't have to set this at all :)

src/components/Proposals/ProposalBanner.js
```diff
@@ -39,43 +42,62 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+            justify-content: space-around;
           `}
         >
-          Aragon Network Cash is a test organization set up as part of a new
-          round of precedence campaign disputes. Learn more on the{' '}
-          <Link href="https://aragon.org/blog/juror-dispute-guide-aragon-network-cash">
-            Aragon project blog
-          </Link>
-          .
+          <div>
```
andy-hook, line 57: Should we use a `LayoutLimiter` to contain this? otherwise at wider sizes things start to expand larger than we'd like:

image

src/components/Proposals/ProposalBanner.js
```diff
@@ -39,43 +42,62 @@ function ProposalBanner({ onCloseBanner, ...props }) {
...
+              .
+            </div>
+          </div>
+
+          <img
+            src={pcBanner}
+            css={`
+              max-width: 90%;
+            `}
+          />
```
andy-hook, line 100: I'm wondering whether we should only render this at larger sizes, or make it smaller on small screens. It takes quite a bit of space on mobile, what do you think?

image

andy-hook (commented):
Looking good! ❤️  I left a few design related comments.

Also wondering, should we keep the old code around to be able to switch easily?

src/components/Proposals/ProposalBanner.js
delfipolito: Oh right!

src/components/Proposals/ProposalBanner.js
delfipolito: You're right, it was odd. Now it looks like this
image

andy-hook:
Also something I've noticed is the graphic rendering is not quite right:

image

vs the design:

image

delfipolito:
You are right! It broke when I run it through the optimiser. Thanks for catching that

andy-hook:
Awesome stuff @delfipolito ! I made a few small tweaks and got the SVG optimised and rendering correctly.

image

andy-hook (approved):

PR #69, opened 2020-09-01
Add Kernel, ACL & Registry icons

delfipolito:
image


src/lib/apps-utils.js
```diff
@@ -1,23 +1,27 @@
+import acl from '../assets/acl.svg'
+import kernel from '../assets/kernel.svg'
+import registry from '../assets/registry.svg'
```
andy-hook, line 3: How about `iconAcl` or similar?

andy-hook (approved):
🙌

Issue #65, opened 2020-09-01
Improve loading performance


andy-hook:
Currently we're doing too much upfront work, that was fine for quickly getting set-up but it's not scalable. There also seems to be quite a bit of added latency with `describeScript` which needs investigating. At this point I think it's worth switching to lazy loading per route and component.

PR #70, opened 2020-09-01
Lazily load votes and agreement

andy-hook:
Fixes https://github.com/aragon/network-dashboard/issues/65

This PR lazily loads the votes and agreement data into separate contexts, and provides loaders in each case so that we can present the app as soon as the organisation is ready.

When we switch to subscribing for updates this will also be useful to persist the subs between views and update data as it comes in.


src/components/Proposals/Proposals.js
```diff
@@ -54,28 +41,58 @@ const Proposals = React.memo(function Proposals() {
...
+
+  const sortedVotes = useMemo(
+    () =>
+      [...votes].sort((a, b) => {
+        const dateDiff = b.endDate - a.endDate
+        // Order by descending voteId if there's no end date difference
+        return dateDiff !== 0 ? dateDiff : b.voteId - a.voteId
+      }),
+    [votes]
+  )
```
andy-hook, line 72: Also sorting the votes in this PR, this was lifted straight from the client as I assume the same applies here.

src/providers/Votes.js
```diff
@@ -0,0 +1,34 @@
+import React, { useContext, useMemo } from 'react'
+import PropTypes from 'prop-types'
+import { useDisputableVotes } from '../hooks/useDisputableVotes'
+
+const VotesContext = React.createContext({
+  votes: [],
+  loading: true,
+})
+
+function VotesProvider({ children }) {
```
andy-hook, line 10: We may end up not need these contexts in the future. I was thinking we could just open a subscription when on a route, and unsub when leaving. I'm a little concerned about down the wire data costs of the polling if we leave everything open, especially on mobile. I think for now this is fine because we're not polling on these yet.

I've added them here because getting the votes at the moment takes a bit of time due to all of the `describeScript` calls, and bouncing between `proposals` and `proposalsSingle` without it repeats all of that work, making it quite irritating from a UX perspective.

andy-hook:
@delfipolito – I'm going to merge this now so I can continue with what I'm working on, please do take a look and comment if anything jumps out.

Issue #72, opened 2020-09-02
Progressively render UI that depends on describeScript


andy-hook:
Calls to connects `describeScript` are slow, we should avoid delaying UI rendering and instead progressively load the descriptions separately inside of our cards and proposal detail page.

PR #74, opened 2020-09-03
Performance: Progressively render descriptions

andy-hook:
Fixes https://github.com/aragon/network-dashboard/issues/72

Improves our perceived performance by unblocking UI rendering and progressively loading each call to `describeScript`, we also now cache those descriptions in a mapping based on `vote.id` so that we don't have to do the same work multiple times.

ezgif-7-354765a93a5a

### DIFF
--- a/src/components/Proposals/Description.js
+++ b/src/components/Proposals/Description.js
@@ -1,6 +1,6 @@
 import React from 'react'
+import { PropTypes } from 'prop-types'
 import { GU, IdentityBadge, useTheme } from '@aragon/ui'
-import { propTypes } from '@aragon/ui/dist/index-46d0e707'
 
 function Description({ path }) {
   return (
@@ -108,7 +108,7 @@ function DescriptionStep({ step }) {
 /* eslint-enable react/prop-types */
 
 Description.propTypes = {
-  path: propTypes.array,
+  path: PropTypes.array,
 }
 
 export default Description
--- a/src/components/Proposals/Description.js
+++ b/src/components/Proposals/Description.js
@@ -1,9 +1,11 @@
 import React from 'react'
 import { GU, IdentityBadge, useTheme } from '@aragon/ui'
+import { propTypes } from '@aragon/ui/dist/index-46d0e707'
 
-/* eslint-disable react/prop-types */
-function Description({ path }) {
-  return (
+function Description({ path, loading }) {
+  return loading ? (
+    'Loading...'
+  ) : (
     <span
       css={`
         // overflow-wrap:anywhere and hyphens:auto are not supported yet by
@@ -21,6 +23,7 @@ function Description({ path }) {
   )
 }
 
+/* eslint-disable react/prop-types */
 function DescriptionStep({ step }) {
   const theme = useTheme()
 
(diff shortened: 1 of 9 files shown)


src/components/Proposals/ProposalCard.js
```diff
@@ -69,12 +79,14 @@ function ProposalCard({ vote, onProposalClick }) {
...
         `}
       >
-        <AppBadge
-          label={target.name ? target.name : target.address}
-          appAddress={target.address}
-          iconSrc={target.icon ? target.icon : ''}
-          badgeOnly
-        />
+        {emptyScript ? (
+          <DefaultAppBadge />
```
andy-hook, line 83: This was based on the previous version and is fallback disputable app badge. @delfipolito Are we sure this should be the default app when there is no description available to process?

src/components/Proposals/ProposalCard.js
```diff
@@ -93,11 +105,17 @@ function ProposalCard({ vote, onProposalClick }) {
           overflow: hidden;
         `}
       >
-        <strong css="font-weight: bold">#{voteId}: </strong>
-        {Array.isArray(description) ? (
-          <Description path={description} />
+        {emptyScript ? (
```
andy-hook, line 108: I don't love this boolean check for an empty script, but it was the cleanest way I could think to present the branching logic depending on what data is available.

src/components/Proposals/ProposalCard.js
```diff
@@ -41,7 +45,13 @@ function getAttributes(status, theme) {
...
 function ProposalCard({ vote, onProposalClick }) {
   const theme = useTheme()
-  const { context, voteId, description, target } = vote
+  const { context, voteId, script } = vote
+  const {
+    description,
+    emptyScript,
+    targetApp,
+    loading: descriptionLoading,
+  } = useDescribeVote(script, vote.id)
```
andy-hook, line 54: This hook also returns the target data for the app badge, this feels a bit cleaner to me and we can always pull it out in the future if we need to use it in other contexts.

src/components/Proposals/Description.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { GU, IdentityBadge, useTheme } from '@aragon/ui'
+import { propTypes } from '@aragon/ui/dist/index-46d0e707'
```
delfipolito, line 3: Should we have  our own proptypes here or export them properly on aragonUI?

src/components/Proposals/ProposalCard.js
delfipolito: Yes :)

src/components/Proposals/Description.js
andy-hook: Ohhh nice spot! my editor auto imported this from the wrong place 🤦  thanks!

src/components/Proposals/ProposalCard.js
andy-hook: Gotcha!

src/components/Proposals/Description.js
andy-hook: Fixed in https://github.com/aragon/network-dashboard/pull/74/commits/f6b10c9365a20c11470edb5595df9e8d67f52e4f

delfipolito (approved):

Issue #19, opened 2020-08-21
Layout: Implement custom two column component


andy-hook:
The existing `<Split/>` component provided by the UI kit has a very narrow secondary column, this works against us at the moment as our designs require more space. It should also let us sidestep [this issue](https://github.com/aragon/aragon-ui/issues/789) on the Proposal detail view.

PR #75, opened 2020-09-03
Global layout: Larger secondary columns, adjustable box

andy-hook:
Fixes https://github.com/aragon/network-dashboard/issues/19

This PR adds custom column layout and adjustable box. This is needed because our UI kit has a lot of opinions about how it would like to present these, and optimises for apps with a sidebar. Our designs need a larger right hand column, and the ability to easily adjust the box padding depending on the context.

I also moved the disputable colouring into the new `LayoutBox` as a `mode` that can be optionally provided.

These changes also fix some overflow blowout we had at smaller viewport sizes.

image

### DIFF
--- a/src/components/Agreement/Agreement.js
+++ b/src/components/Agreement/Agreement.js
@@ -1,11 +1,13 @@
 import React from 'react'
-import { Box, Button, Header, Split } from '@aragon/ui'
+import { Button, Header } from '@aragon/ui'
 import LayoutGutter from '../Layout/LayoutGutter'
 import LayoutLimiter from '../Layout/LayoutLimiter'
 import AgreementBindingActions from './AgreementBindingActions'
 import AgreementHeader from './AgreementHeader'
 import AgreementDetails from './AgreementDetails'
 import AgreementDocument from './AgreementDocument'
+import LayoutBox from '../Layout/LayoutBox'
+import LayoutColumns from '../Layout/LayoutColumns'
 import LoadingSection from '../Loading/LoadingSection'
 import { useAgreement } from '../../providers/Agreement'
 
@@ -40,10 +42,10 @@ function AgreementLayout({ agreementDetails }) {
   } = agreementDetails
 
   return (
-    <Split
+    <LayoutColumns
       primary={
         <>
-          <Box>
+          <LayoutBox primary>
             <AgreementHeader title={title} />
             <AgreementDetails
               contractAddress={contractAddress}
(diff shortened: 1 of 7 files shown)


src/components/Agreement/Agreement.js
```diff
@@ -40,18 +42,18 @@ function AgreementLayout({ agreementDetails }) {
   } = agreementDetails
 
   return (
-    <Split
+    <LayoutColumns
       primary={
         <>
-          <Box>
+          <LayoutBox primary>
```
andy-hook, line 48: I've been prefixing our custom layout components with `Layout`. This makes the distinction clear vs using the same names as exist in the UI kit.

The `primary` prop is what gives the box extra padding for use in the `primary` slot. Previously the box would inherit this implicitly, which caused problems such as those [described here](https://github.com/aragon/network-dashboard/issues/19)

src/components/Layout/LayoutBox.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      negative: {
+        backgroundColor: '#FFFAFA',
+        borderColor: '#FF7C7C',
+      },
+      disabled: {
+        backgroundColor: theme.surfacePressed,
+        borderColor: theme.border,
+      },
+    }
```
andy-hook, line 24: I'm using semantic naming here to keep this component generalised, the disputable states can then be easily mapped to use them.

src/components/Layout/LayoutColumns.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { useLayout, GU } from '@aragon/ui'
+
+function LayoutColumns({ primary, secondary }) {
+  const { layoutName } = useLayout()
+  const oneColumn = layoutName === 'small' || layoutName === 'medium'
```
delfipolito, line 7: I know I did it like this the other time. But I just realised that here we can also use useViewport hook and do it this way:  `const oneColumn = below('medium')`. Both are fine :) whatever you prefer.

delfipolito (approved):

delfipolito (approved):
This looks awesome!! So much better 🤩 🤩 🤩

src/components/Layout/LayoutColumns.js
andy-hook: That's definitely nicer, In the future i'd like to useViewport for all of these types of things, the problem is it's not set-up with our breakpoints yet and a lot of the UI kit still relies on useLayout. I was thinking we should continue to useLayout for the time being and switch it all over together :)

PR #76, opened 2020-09-03
Proposals: Fix card tabbing

andy-hook:
Badges in descriptions were stealing focus in the card list, this improves the tabbing experience.
Also fixes a nested dom warning introduced in https://github.com/aragon/network-dashboard/pull/74

### DIFF
--- a/src/components/Proposals/Description.js
+++ b/src/components/Proposals/Description.js
@@ -2,7 +2,7 @@ import React from 'react'
 import { PropTypes } from 'prop-types'
 import { GU, IdentityBadge, useTheme } from '@aragon/ui'
 
-function Description({ path }) {
+function Description({ disableBadgeInteraction, path }) {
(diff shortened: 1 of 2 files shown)


src/components/Proposals/Description.js
```diff
@@ -108,6 +115,7 @@ function DescriptionStep({ step }) {
 /* eslint-enable react/prop-types */
 
 Description.propTypes = {
+  disableBadgeInteraction: PropTypes.bool,
```
delfipolito, line 118: Should we put false as default in this prop?

delfipolito (approved):

src/components/Proposals/Description.js
andy-hook: Good point, though it's falsy by default, being explicit wouldn't hurt.

PR #77, opened 2020-09-04
Sentry: Provide logging via utils

andy-hook:
Little cleanup ahead of some error UX related work.


andy-hook:
@delfipolito Going to merge this now to keep going with what I'm working on, not a big change and shouldn't cause you any issues :)

PR #78, opened 2020-09-04
Account module

delfipolito:
screen


src/components/Account/AccountModule.js
```diff
@@ -0,0 +1,152 @@
...
+      // reject closing the popover
+      return false
+    }
+    setOpened(false)
+  }, [screenId])
+
+  return (
+    <div
+      ref={buttonRef}
+      tabIndex="0"
```
andy-hook, line 83: Do we need this? it seems to interfere with keyboard tabbing, but I'm curious if it exists to fix some other issue 🤔

package.json
```diff
@@ -49,7 +50,8 @@
     "react-router-dom": "^5.2.0",
     "react-spring": "^8.0.1",
     "styled-components": "^5.1.1",
-    "use-viewport": "^0.2.0"
+    "use-viewport": "^0.2.0",
+    "use-wallet": "^0.8.0"
```
andy-hook, line 54: Let's make sure to commit an updated lockfile for these new dependencies :)

andy-hook (commented):
Looking super solid! 🚀 🚀 my main comments are about the network configuration which we already have available :)

src/components/Account/AccountButton.js
```diff
@@ -0,0 +1,102 @@
...
+  EthIdenticon,
+  IconDown,
+  GU,
+  RADIUS,
+  textStyle,
+  useTheme,
+  useLayout,
+} from '@aragon/ui'
+import { useWallet } from 'use-wallet'
+import { shortenAddress } from '../../lib/web3-utils'
```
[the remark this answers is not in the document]
delfipolito: Thats supper weird! I swear it was there haha. I added it now :) thanks for the catch

src/components/Account/ScreenConnected.js
```diff
@@ -80,13 +80,12 @@ function AccountScreenConnected({ wallet }) {
...
             <IdentityBadge
               entity={wallet.account}
-              label={shortenAddress(wallet.account)}
               compact
               badgeOnly
               css={`
                 cursor: pointer;
               `}
-              networkType={getNetworkType()}
+              networkType={getNetworkName()}
```
andy-hook: We should pass the `chainId` in here to get the name

src/lib/web3-utils.js
```diff
@@ -62,3 +62,17 @@ export function addressesEqual(first, second) {
   return first === second
 }
 
+export function shortenAddress(address, charsLength = 4) {
```
andy-hook, line 53: Let's add the docs annotation for this util as we already have one available

yarn.lock
```diff
@@ -64,6 +64,15 @@
     "@aragon/connect-ethereum" "^0.7.0-beta.3"
     "@aragon/connect-thegraph" "^0.7.0-beta.3"
 
+"@aragon/provided-connector@^6.0.7":
```
andy-hook, line 67: Just checking – Are we definitely updating the existing lock file? the diff just looks a little large so want to make sure!

andy-hook (commented):
A couple more little things ✨  I'm also going to check what's happening with the bundle size, seems a bit large.

yarn.lock
andy-hook: Never mind, use-wallet seems to be bring plenty of friends 😅

delfipolito: yes :)

src/components/Account/ScreenConnected.js
delfipolito: This is outdated, now the code looks like this: `networkType={networkEnvironment.legacyNetworkType}` :)

andy-hook (approved):
🚀 🚀 🚀

PR #79, opened 2020-09-04
Connect: Use vanilla library

andy-hook:
This moves us from `connect-react` to the vanilla library. I've done this primarily to give us the flexibility to workaround any potential issues that might occur in the short term, plus a couple of small issues i'm experiencing. Ideally we will switch back to the react version once things have stabilised.


andy-hook:
Closing for now as other PRs are higher priority, I always want to triple check what I need to do can't be done with `connect-reat` after a new release.

PR #80, opened 2020-09-04
Error handling: Add generic boundary

andy-hook:
Add feedback for generic errors and when navigating to out of bounds proposalIds

image


src/components/Errors/GenericError.js
```diff
@@ -0,0 +1,145 @@
...
+                  >
+                    {detailsTitle}
+                  </h2>
+                )}
+                <p
+                  css={`
+                    color: ${theme.surfaceContentSecondary};
+                  `}
+                >
+                  {detailsContent}
```
delfipolito, line 115: is detsailsContent always going to be present?

src/components/Errors/GenericError.js
```diff
@@ -0,0 +1,145 @@
...
+        </div>
+      )}
+      <div
+        css={`
+          display: grid;
+          grid-gap: ${1 * GU}px;
+          grid-template-columns: auto auto;
+        `}
+      >
+        <Button onClick={reportCallback}>Send Report</Button>
```
delfipolito, line 129: Maybe we should make this function required in propTypes,  or render the button only if it is present

delfipolito (approved):
👏🏻👏🏻

src/components/Errors/GenericError.js
andy-hook: Yep, there'll always be a stack provided by an error.

src/components/Errors/GenericError.js
andy-hook: Good spot! this should only be rendered if provided.

PR #87, opened 2020-09-07
Account Module: Minor adjustment

andy-hook:
Address the issue we discussed with "Unknown" wallet during connection. Couple of other minor tweaks to click targets and header spacing while I was in the area.

### DIFF
--- a/src/components/Account/AccountModule.js
+++ b/src/components/Account/AccountModule.js
@@ -28,20 +28,13 @@ function AccountModule() {
   const toggle = useCallback(() => setOpened((opened) => !opened), [])
 
   useEffect(() => {
-    let timer
-
     if (status === 'error') {
       setActivatingDelayed(null)
     }
 
     if (status === 'connecting') {
       setActivatingDelayed(connector)
-      timer = setTimeout(() => {
-        setActivatingDelayed(null)
-      }, 400)
     }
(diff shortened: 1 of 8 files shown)


src/components/Account/AccountModule.js
```diff
@@ -28,20 +28,13 @@ function AccountModule() {
   const toggle = useCallback(() => setOpened((opened) => !opened), [])
 
   useEffect(() => {
-    let timer
```
delfipolito, line 31: Just curious, why are we deleting this?

delfipolito (approved):
Looks good to me!

src/components/Account/AccountModule.js
andy-hook: This was added as workaround for an animation issue [here](https://github.com/aragon/court-dashboard/pull/375/files#r463716108) and setting this value to `null` after the timeout is what makes the module unable to determine the type of wallet that is connecting. I gave it a good amount of testing and saw no animation issues without it.

delfipolito: awesome! :)

andy-hook: But keep an eye out, if all is well we can suggest the same changes in our other repos :)

PR #88, opened 2020-09-08
Card accordion

delfipolito:
card accordion


src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+  GU,
+  IconUp,
+  IconDown,
+  RADIUS,
+  useTheme,
+} from '@aragon/ui'
+import { Spring, Transition, animated } from 'react-spring/renderprops'
+import styled from 'styled-components'
+
+const CardAccordion = React.memo(function CardAccordion({ card, expansion }) {
```
andy-hook, line 15: We probably don't need to memoize this component as the parent is taking care of it. Also, how do you feel about a different component name? Accordion to me generally feels like a term for multiple elements in a list. What do you think of `ExpandableCard`?

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+    >
+      <OpenedSurfaceBorder opened={opened} />
+      <Card
+        css={`
+          width: 100%;
+          z-index: 2;
+        `}
+      >
+        <ToggleButton onClick={toggleButton} opened={opened} />
+        {card}
```
andy-hook, line 38: How'd you feel about a different prop name for this, something like `content`?

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+        css={`
+          margin-top: ${-RADIUS}px;
+        `}
+      >
+        <Transition
+          native
+          items={opened}
+          from={{ height: 0 }}
+          enter={{ height: 'auto' }}
+          leave={{ height: 0 }}
```
andy-hook, line 50: This works great now! 👏

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+          }
+        </Transition>
+      </div>
+    </div>
+  )
+})
+
+const Expansion = styled(animated.div)`
+  z-index: 1;
+`
```
andy-hook, line 84: This is fine, but I think we've generally been moving away from separated styled components. What do you think about declaring a single `AnimatedDiv` and moving this inline?

```javascript
<AnimatedDiv style={props} css={`...customcss`}>
```

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+        margin-bottom: ${RADIUS}px;
+      `}
+    >
+      <OpenedSurfaceBorder opened={opened} />
+      <Card
+        css={`
+          width: 100%;
+          z-index: 2;
+        `}
+      >
```
andy-hook, line 36: At the moment these cards seem to have a fixed height. I was thinking it would be better to set it to `auto` and use padding to wrap the inner content. We could then aim to match the height in the design:

image

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+  onClick: PropTypes.func,
+}
+
+function OpenedSurfaceBorder({ opened }) {
+  return (
+    <Spring
+      native
+      from={{ width: 0 }}
+      to={{ width: Number(opened) }}
+      config={{ mass: 0.8, tension: 300, friction: 28 }}
```
andy-hook, line 143: Is there anything in our existing springs that would work nicely here? you can check in `styles/springs`
No worries if not, but we could move it there too if you think it feels unique and could be useful in other places!

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+    >
+      {({ width }) => (
+        <Surface
+          style={{
+            transform: width.interpolate((v) => `scale3d(${v}, 1, 1)`),
```
andy-hook, line 148: High performance transform, nice! 🚀

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,173 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  height: 100%;
+  width: 3px;
+  border-top-left-radius: ${RADIUS}px;
+  border-bottom-left-radius: ${RADIUS}px;
+  background: linear-gradient(90deg, #32fff5 -103.98%, #01bfe3 80.13%);
+  transform-origin: 0 0;
+`
```
andy-hook, line 171: Similar to a previous comment, let's inline this and reuse our `AnimatedDiv`!

src/components/Stake/SideBar.js
```diff
@@ -0,0 +1,105 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { GU, textStyle, useTheme } from '@aragon/ui'
+import CardAccordion from './CardAccordion'
+import coin from './assets/coin.svg'
+import wallet from './assets/wallet.svg'
+
+const Sidebar = React.memo(function Sidebar() {
```
andy-hook, line 8: Probably don't need to memoize just yet, if we feel we need it later for performance we can always come back to it!

src/components/Stake/SideBar.js
```diff
@@ -0,0 +1,105 @@
...
+      <CardAccordion
+        card={
+          <CardContent
+            amount="72,243.47"
+            icon={wallet}
+            title="Available ANT"
+            tokenAmount="11,278.04"
+          />
+        }
+        expansion="I'm disputing this proposal because I believe that it was not made in good faith and will not benefit all ANT holders in equal measure. I'm disputing this proposal because I believe that it was not made in good faith and will not benefit all ANT holders in equal measure. ANT holders in equal measure."
```
andy-hook, line 27: Do we know what the content of these cards is going to be yet? perhaps worth checking in with @dizzypaty as we'll need to have these ready to go soon.

src/components/Stake/SideBar.js
```diff
@@ -0,0 +1,105 @@
...
+  return (
+    <div
+      css={`
+        text-align: center;
+      `}
+    >
+      <img
+        src={icon}
+        width={6 * GU}
+        height={6 * GU}
```
andy-hook, line 67: Nice one for reserving space while the images load! ✨

src/components/Stake/SideBar.js
```diff
@@ -0,0 +1,105 @@
...
+        css={`
+          margin: ${0.5 * GU}px 0;
+        `}
+      >
+        {title}
+      </h1>
+      <p
+        css={`
+          color: ${negative ? theme.negative : theme.positive};
+        `}
```
andy-hook, line 89: What would be super cool is if we set the `color` of this text depending on whether the `amount` is positive or negative, then we wouldn't need this prop at all, what do you think?

src/components/Stake/Stake.js
```diff
@@ -0,0 +1,28 @@
...
+import LayoutGutter from '../Layout/LayoutGutter'
+import LayoutLimiter from '../Layout/LayoutLimiter'
+import SideBar from './SideBar'
+
+const Stake = React.memo(function Stake() {
+  return (
+    <LayoutGutter
+      css={`
+        position: relative;
+        z-index: 2;
```
andy-hook, line 14: Just checking whether we need these values here, z-index should already be being handled higher in the tree.

src/components/Stake/Stake.js
```diff
@@ -0,0 +1,28 @@
...
+      `}
+    >
+      <LayoutLimiter>
+        <Header primary="Stake management" />
+        <LayoutColumns
+          primary={<LayoutBox primary />}
+          secondary={<SideBar />}
+        />
```
andy-hook, line 22: I've added an issue for an improvement to allow us to invert the columns [here](https://github.com/aragon/network-dashboard/issues/90) feel free to add that improvement separately if you like!

src/components/Stake/assets/coin.svg
```diff
@@ -0,0 +1,45 @@
...
+</g>
+<path opacity="0.7" fill-rule="evenodd" clip-rule="evenodd" d="M40.0065 27.4111C39.8512 27.4111 39.683 27.4637 39.492 27.5717C39.2101 27.7313 38.8541 27.8538 38.4772 27.9835L38.3679 28.0212C36.3102 28.7327 33.6152 29.1246 30.7794 29.1246C27.9436 29.1246 25.2486 28.7327 23.1909 28.0212C22.609 27.82 22.4472 27.7568 22.2322 27.6463C22.0136 27.5317 21.8102 27.4735 21.6274 27.4735C21.2313 27.4735 20.9881 27.7393 20.9226 28.0026C20.8921 28.1254 20.8767 28.2506 20.8767 28.3749C20.8767 29.4236 21.956 30.3743 23.9158 31.0519C25.7546 31.6876 28.1921 32.0378 30.7793 32.0378C33.3666 32.0378 35.804 31.6876 37.6429 31.0519C39.6027 30.3743 40.682 29.4236 40.682 28.3749C40.682 28.2117 40.6555 28.0477 40.6023 27.8854C40.4672 27.4728 40.1711 27.4111 40.0065 27.4111Z" fill="url(#paint1_linear)"/>
```
andy-hook, line 6: Let's optimise these assets!

andy-hook (commented):
Great stuff! 💯 🏆  I've added some comments but they are mostly small little details, style changes and the odd suggestion, nothing major. Nice work with the animation!

src/components/Stake/assets/wallet.svg
```diff
@@ -0,0 +1,15 @@
+<svg width="49" height="49" viewBox="0 0 49 49" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M34.8086 24.0064V21.8636C34.8086 21.0745 34.137 20.435 33.3086 20.435H16.8086C15.9813 20.435 15.3086 19.7943 15.3086 19.0064C15.3086 18.2186 15.9813 17.5778 16.8086 17.5778H33.3086C34.137 17.5778 34.8086 16.9382 34.8086 16.1493C34.8086 15.3603 34.137 14.7207 33.3086 14.7207H16.8086C14.3271 14.7207 12.3086 16.6432 12.3086 19.0064V20.435V31.8636C12.3086 33.4415 13.6518 34.7207 15.3086 34.7207H33.3086C34.137 34.7207 34.8086 34.0812 34.8086 33.2921V31.1493C35.637 31.1493 36.3086 30.5097 36.3086 29.7207V25.435C36.3086 24.646 35.637 24.0064 34.8086 24.0064ZM32.5586 29.0064C31.7302 29.0064 31.0586 28.3669 31.0586 27.5778C31.0586 26.7888 31.7302 26.1493 32.5586 26.1493C33.387 26.1493 34.0586 26.7888 34.0586 27.5778C34.0586 28.3669 33.387 29.0064 32.5586 29.0064Z" fill="url(#paint0_linear)"/>
```
andy-hook, line 2: Something I noticed with these icons is that I think the cropping is a little off in our exports. The coins tend to render smaller than the wallet, might be worth checking them :)

image

src/components/Stake/Stake.js
delfipolito: yes! I was planning to in the next stake management PR :)

src/components/Stake/SideBar.js
delfipolito: Do you mean the expansion? Yes, she is on it :)

src/components/Stake/SideBar.js
delfipolito: yeah! sounds good! I will address this comment when I have the real data to see how it looks like before implementing it :)

src/components/Stake/Stake.js
delfipolito: good catch! this wasnt meant to be here

package.json
```diff
@@ -45,7 +45,7 @@
     "ethers": "^5.0.8",
     "is-ipfs": "^2.0.0",
     "prop-types": "^15.7.2",
-    "react": "^16.13.1",
+    "react": "^16.12.0",
```
andy-hook, line 48: Not sure what happened with this version change, but we should keep the current :)

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,177 @@
...
+} from '@aragon/ui'
+import { Spring, Transition, animated } from 'react-spring/renderprops'
+
+const AnimatedDiv = animated.div
+
+function CardAccordion({ content, expansion }) {
+  const [opened, setOpened] = useState(false)
+  const theme = useTheme()
+
+  const toggleButton = useCallback((index) => {
```
andy-hook, line 21: Do we need `index` here?

src/components/Stake/CardAccordion.js
```diff
@@ -0,0 +1,177 @@
...
+  IconDown,
+  RADIUS,
+  springs,
+  useTheme,
+} from '@aragon/ui'
+import { Spring, Transition, animated } from 'react-spring/renderprops'
+
+const AnimatedDiv = animated.div
+
+function CardAccordion({ content, expansion }) {
```
andy-hook, line 17: Did we want to rename to `ExpandableCard`? not a big deal though!

andy-hook (approved):
Looks and works awesome! Two more tiny tiny comments but otherwise LGTM 🎉

PR #91, opened 2020-09-08
Add settled state

delfipolito:


package.json
```diff
@@ -35,7 +35,7 @@
   },
   "dependencies": {
     "@aragon/connect-agreement": "^0.0.0-beta.4",
```
andy-hook, line 37: Should we be upgrading agreement as well?

delfipolito: I didnt know if it would break something, that's why i didnt do it on this PR

delfipolito: But yeah! Should I do it in this one?

andy-hook: I'm just going to check a few things, I'm seeing errors on mainnet with this change.

andy-hook (approved):
We're in business 🙌

PR #94, opened 2020-09-09
Add stepper component

andy-hook:
Most of this is based on my previous client implementation. Error and success info will be added separately.

ezgif-7-198435572f61


andy-hook:
Code has been extensively reviewed in multiple past iterations, I'm going to merge and will be happy to address additional comments separately.

PR #95, opened 2020-09-09
Quiet ending info

delfipolito:
- Show quiet ending period only when vote is active
- Show quiet ending extension when the vote gets extended


src/components/Proposals/ProposalDetails/DisputableActionStatus.js
```diff
@@ -60,6 +65,34 @@ function DisputableActionStatus({ vote }) {
...
+            </span>
+            <span
+              css={`
+                color: ${theme.surfaceContentSecondary};
+              `}
+            >
+              Hours
+            </span>
+          </Item>
+        )}
```
andy-hook, line 95: Just checking, are these ordered correctly? The designs have them above the info box

image

andy-hook (commented):
Nice one for adding these! something that i'm wondering is that these conditionals might soon become difficult to manage and follow. While it's been fine until now, I think we're at the point where we need to extract logic and consider our approach: e.g. have a clear understanding of all of our possible states, make them easy to follow and map things like the descriptive info from those states.

I think hooks could help us here, we probably don't need to do this right away but the next time we come back here to add more we should think about reworking how some of the logic is structured.

src/components/Proposals/ProposalDetails/DisputableActionStatus.js
delfipolito: you are right! they should be after the timers

src/components/Proposals/ProposalDetails/DisputableActionStatus.js
```diff
@@ -6,17 +6,22 @@ import {
...
+  const theme = useTheme()
   const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
   const challenged = disputableStatus === VOTE_STATUS_PAUSED
+  const scheduled = disputableStatus === VOTE_STATUS_ACTIVE
   const challengeEndDate = toMs(vote.challengeEndDate)
   const pausedAt = toMs(vote.pausedAt)
   const voteEndDate = toMs(vote.endDate)
+  const extendedPeriod =
+    vote.quietEndingExtensionDuration &&
+    parseInt(vote.quietEndingExtensionDuration) > 0
```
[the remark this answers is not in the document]
delfipolito: I guarded it cause  i  wasnt sure if this variable is always gonna come, but I can check  with Facu to delete  this :)

andy-hook (approved):
Thanks for making the changes! LGTM

PR #96, opened 2020-09-09
Staking balance card

delfipolito:
- Balance carrd
- Inverted mode on LayoutColumns

image


src/components/Stake/Balance.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { Button, Card, GU, textStyle, useTheme } from '@aragon/ui'
+import tokenIcon from './assets/tokenIcon.svg'
+
+function Balance() {
```
andy-hook, line 5: How do you feel about `BalanceCard` ?

src/components/Stake/Balance.js
```diff
@@ -0,0 +1,62 @@
...
+  const theme = useTheme()
+
+  return (
+    <Card
+      css={`
+        width: 100%;
+        z-index: 2;
```
andy-hook, line 12: Do we need the `z-index`?

andy-hook (approved):
Couple of little nits but otherwise LGTM!

PR #97, opened 2020-09-10
Proposal: Use connected account

andy-hook:


src/components/Proposals/ProposalDetails/DisputableActions.js
```diff
@@ -46,9 +46,9 @@ function DisputableActions({ status, submitter }) {
       <>
         <Button
           mode="strong"
-          disabled
-          wide
           label="Challenge proposal"
+          disabled={!connectedAccount}
```
delfipolito, line 50: I think we should keep this button disabled (despite we have the real connectedAccount) until it can perform the action

delfipolito (commented):
Nice work removing TODO's  🙌🏼

src/components/Proposals/ProposalDetails/DisputableActions.js
andy-hook: Sounds good!

delfipolito (approved):

PR #98, opened 2020-09-10
Signing modals: Add scaffold, agreement signing

andy-hook:
I've built out the bones of these 3 modals as a guide for adding the signing functionality for the three features we've been working on. There is some boilerplate here but these modals will become quite complex and the ability to encapsulate state per step in the modal, as well as across screens is important.

The multimodal might look a little strange but it's necessary to make sure state gets cleaned up consistently and correctly between open and closing:
```javascript
<MultiModal>
     <MultiModalScreens/>
</MultiModal>
```

As far as I can tell we haven't used modals for this amount of interactions before, especially with this many steps, so I had to do some magic with `use-inside` to make the composition a bit nicer.

Sorry about the big modal diff, it's just a restructure rather than a big code change.


src/components/ModalFlows/ChallengeProposalModal/ChallengeProposalModal.js
```diff
@@ -0,0 +1,44 @@
...
+function ChallengeProposalModal() {
+  const mounted = useMounted()
+  const [transactions, setTransactions] = useState([])
+
+  const challengeProposal = useCallback(
+    async (onComplete) => {
+      try {
+        if (mounted()) {
+          setTransactions([])
+        }
```
andy-hook, line 16: We can set the transactions here after receiving them via the connector, they then get passed to the Stepper in the next screen which will prompt the signing/s

src/components/ModalFlows/TransactionStepper.js
```diff
@@ -0,0 +1,69 @@
...
+import Stepper from '../Stepper/Stepper'
+import { useWallet } from '../../providers/Wallet'
+
+const indexNumber = {
+  0: 'First',
+  1: 'Second',
+  2: 'Third',
+  3: 'Fourth',
+  4: 'Fifth',
+}
```
andy-hook, line 13: This is temporary, ideally we will provide step titles via connect data in the future.

src/components/Proposals/ProposalDetails/ProposalDetails.js
```diff
@@ -105,7 +106,10 @@ function ProposalDetails({ vote }) {
       }
       secondary={
         <>
-          <DisputableActionStatus vote={vote} />
+          <DisputableActionStatus
+            vote={vote}
+            disputableStatus={disputableStatus}
```
delfipolito, line 111: Why are we passing  this new prop? Are we using it on the component?

delfipolito (approved):
Hey! This is looking amazing!!!
My only comment is that we should disable the sign agreement button when the user has already sign it (if its possible) 
And also if  possible,  maybe we can put an error  description if an error occurs.

But is looking  great! We can take care of that on another PR

src/components/Proposals/ProposalDetails/ProposalDetails.js
andy-hook: This is a mistake 🙈

andy-hook:
> Hey! This is looking amazing!!!
> My only comment is that we should disable the sign agreement button when the user has already sign it (if its possible)
> And also if possible, maybe we can put an error description if an error occurs.
> 
> But is looking great! We can take care of that on another PR

We discussed offline, but to clarify here. Yes totally agree, was waiting on come connect changes to know if users have signed the agreement, and better errors will come in a separate improvement.

PR #111, opened 2020-09-14
Vote buttons 

delfipolito:
Different scenarios for the vote buttons. 
Is still work in progress because it is missing data from the connector


src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,220 @@
...
+  RADIUS,
+  textStyle,
+  useTheme,
+} from '@aragon/ui'
+import { dateFormat, toMs } from '../../../utils/date-utils'
+
+import { useWallet } from '../../../providers/Wallet'
+
+export const VOTE_YEA = 'VOTE_YEA'
+export const VOTE_NAY = 'VOTE_NAY'
```
andy-hook, line 20: Just wondering about the export here, are we intending to use these outside of this component?

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,220 @@
...
+            </Info>
+          </>
+        )}
+      </>
+    )
+  }
+
+  if (canUserVote && hasVoted && !changeVote) {
+    return (
+      <div>
```
andy-hook, line 60: Do these have layout that requires div, or can they be fragments?

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,220 @@
...
+        began ({dateFormat(toMs(startDate))}) and therefore cannot participate
+        in this vote. Make sure your accounts are holding{' '}
+        <strong>{tokenSymbol}</strong> at the time a vote begins if you'd like
+        to vote using this Voting app.
+      </Info>
+    </div>
+  )
+})
+
+const Buttons = ({ onClickYes = noop, onClickNo = noop, disabled = false }) => (
```
andy-hook, line 153: We can remove default noops if they are provided as defaultProps to the component :)

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,220 @@
...
+      />
+      No
+    </VotingButton>
+  </ButtonsContainer>
+)
+
+const ButtonsContainer = styled.div`
+  display: flex;
+  margin-bottom: ${2 * GU}px;
+`
```
andy-hook, line 179: Let's inline this as we only use it once.

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,220 @@
...
+import { dateFormat, toMs } from '../../../utils/date-utils'
+
+import { useWallet } from '../../../providers/Wallet'
+
+export const VOTE_YEA = 'VOTE_YEA'
+export const VOTE_NAY = 'VOTE_NAY'
+const EMPTY_SCRIPT = '0x00000001'
+
+/* eslint-disable react/prop-types */
+const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
```
andy-hook, line 24: We should provide propTypes for this top level  component, feel free to disable the controlled props on internal elements though!

andy-hook (commented):
I know it's a WIP but wanted to add some comments that might help ahead of time!

Let me know what you think :)

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,251 @@
...
+    const theme = useTheme()
+    const { account } = useWallet()
+    const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
+    const { snapshotBlock, startDate, hasEnded, voterInfo, orgToken } = vote
+
+    if (disputableStatus !== VOTE_STATUS_SCHEDULED) {
+      return <></>
+    }
+
+    if (!account || !voterInfo) {
```
andy-hook, line 33: I think just checking for the account would be enough for this block

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,251 @@
...
+              margin-top: ${2 * GU}px;
+            `}
+          >
+            Connect to your Ethereum provider by clicking on the{' '}
+            <strong
+              css={`
+                display: inline-flex;
+                align-items: center;
+                position: relative;
+                top: 7px;
```
andy-hook, line 63: Let's use `GU` units

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,251 @@
...
+            background: ${theme.background};
+            padding: ${3.5 * GU}px ${10 * GU}px;
+            text-align: center;
+          `}
+        >
+          <div
+            css={`
+              ${textStyle('body1')};
+            `}
+          >
```
andy-hook, line 47: Let's use paragraphs for the text in this block

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,251 @@
...
+      )
+    }
+
+    if (
+      !hasEnded &&
+      voterInfo.hasVoted &&
+      parseInt(voterInfo.accountBalance) > 0
+    ) {
+      return (
+        <div>
```
andy-hook, line 100: Can this be a fragment?

src/components/Proposals/ProposalDetails/VoteActions.js
```diff
@@ -0,0 +1,251 @@
...
+            <strong>change your vote</strong> as many times as you wish.
+          </Info>
+        </div>
+      )
+    }
+
+    if (voterInfo.canVote) {
+      return (
+        <>
+          <Buttons onClickYes={onVoteYes} onClickNo={onVoteNo} />
```
andy-hook, line 122: Let's match the props here for clarity, I think `onVoteYes` is the better option

src/components/Proposals/ProposalDetails/ProposalDetails.js
```diff
@@ -65,8 +66,7 @@ function ProposalDetails({ vote }) {
     return disputablePresentation[disputableStatus] || {}
   }, [disputableStatus])
 
-  // TODO: get youVoted flag from connector
-  const youVoted = false
+  const youVoted = voterInfo && voterInfo.hasVoted
```
andy-hook, line 69: `accountHasVoted`?

src/components/Proposals/ProposalDetails/ProposalDetails.js
```diff
@@ -104,6 +104,14 @@ function ProposalDetails({ vote }) {
               vote={vote}
               disabledProgressBars={disabledProgressBars}
             />
+            {youVoted && (
+              <VoteCast
+                accountVote={voterInfo.hasVoted}
```
andy-hook, line 109: I think it would be simpler to make this prop a boolean and pass `accountHasVoted` with a matching prop name.

src/components/Proposals/ProposalDetails/VoteCast.js
```diff
@@ -0,0 +1,93 @@
...
+          >
+            You voted{' '}
+            <span
+              css={`
+                color: ${theme.surfaceContent};
+                font-weight: 600;
+                text-transform: uppercase;
+              `}
+            >
+              {accountVote === VOTE_YEA ? 'yes' : 'no'}
```
andy-hook, line 68: As mentioned in a previous comment, we could pass this prop as a boolean and check 

```javascript
{accountHasVoted ? 'yes' : 'no'}
```

Then we could remove the `VOTE_YEA` string comparison

src/components/Proposals/ProposalDetails/VoteCast.js
```diff
@@ -0,0 +1,93 @@
...
+            .
+          </p>
+        </div>
+      </div>
+    </div>
+  )
+}
+
+VoteCast.propTypes = {
+  accountVote: PropTypes.node,
```
andy-hook, line 88: Should make this a bool per previous :)

src/hooks/useDisputableVotes.js
```diff
@@ -61,18 +64,33 @@ export function useDisputableVote(proposalId) {
...
+        if (account) {
+          const balance = await token.balance(account)
+          voterInfo = {
+            accountBalanceNow: formatTokenAmount(balance, orgToken.decimals),
+            accountBalance: await vote.formattedVotingPower(account),
+            hasVoted: await vote.castVote(account),
+            canExecute: await vote.canExecute(account),
+            canVote: await vote.canVote(account),
+          }
+        }
```
andy-hook, line 85: Can we combine some of these with `Promise.all`? Don't worry too much about this if it's not too slow. I'd like to update this code shortly anyway with the polling changes.

src/components/Proposals/ProposalDetails/VoteCast.js
delfipolito: here we are not checking if  the account has voted, but if the vote is  "yes" or "no". thats why voterInfo.hasVoted is not a boolean. It can be:  null,  'VOTE_YEA', 'VOTE_NO'

src/components/Proposals/ProposalDetails/VoteActions.js
delfipolito: Is not enough cause first the account is connected and then the promises for voterInfo get executed. And if voterInfo is still undefined, it breaks

andy-hook (approved):

PR #112, opened 2020-09-14
Status info for settled and disputed votes

delfipolito:
image

image

### DIFF
--- a/src/components/Proposals/ProposalDetails/StatusInfo.js
+++ b/src/components/Proposals/ProposalDetails/StatusInfo.js
@@ -11,83 +11,62 @@ import {
 
 function StatusInfo({ vote }) {
   const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
-  if (disputableStatus === VOTE_STATUS_CHALLENGED) {
-    return (
-      <Info
-        mode="warning"
-        css={`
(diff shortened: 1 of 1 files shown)


src/components/Proposals/ProposalDetails/StatusInfo.js
```diff
@@ -24,6 +26,37 @@ function StatusInfo({ vote }) {
...
+        mode="warning"
+        css={`
+          margin-top: ${2 * GU}px;
+        `}
+      >
+        This vote has been canceled as the result of the originating action
+        being challenged and the settlement offer being accepted.
+      </Info>
+    )
+  }
```
andy-hook, line 42: We should create a mapping for these descriptions based on the status, e.g:
```javascript
const infoDescription = {
    [VOTE_STATUS_SETTLED]: 'This vote has been canceled as the result of the originating action
        being challenged and the settlement offer being accepted.',
    [VOTE_STATUS_DISPUTED]: '....',
}
```

And then our template can be as simple as:

```javascript
<Info
     mode="warning"
     css={`
       margin-top: ${2 * GU}px;
     `}
   >
     {infoDescription[disputableStatus]}
   </Info>
```

andy-hook (approved):
🙌

PR #114, opened 2020-09-14
App badges in details screen

delfipolito:
Put TargetAppBage in a separate component so it can be used in the voote cards and in the details screen

image

### DIFF
--- a/src/components/Proposals/ProposalDetails/ProposalDetails.js
+++ b/src/components/Proposals/ProposalDetails/ProposalDetails.js
@@ -39,7 +39,7 @@ import LoadingSkeleton from '../../Loading/LoadingSkeleton'
 import { useWallet } from '../../../providers/Wallet'
 
 function ProposalDetails({ vote }) {
-  const { voteId } = vote
+  const { voteId, id, script } = vote
   const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
 
   const { boxPresentation, disabledProgressBars } = useMemo(() => {
@@ -86,7 +86,7 @@ function ProposalDetails({ vote }) {
                 justify-content: space-between;
               `}
             >
-              <TargetAppBadge vote={vote} />
+              <TargetAppBadge script={script} voteId={id} />
               {youVoted && (
                 <Tag icon={<IconCheck size="small" />} label="Voted" />
               )}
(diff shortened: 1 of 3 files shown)


src/components/Proposals/ProposalDetails/ProposalDetails.js
```diff
@@ -79,15 +80,17 @@ function ProposalDetails({ vote }) {
...
+                display: flex;
+                justify-content: space-between;
+              `}
+            >
+              <TargetAppBadge vote={vote} />
+              {youVoted && (
                 <Tag icon={<IconCheck size="small" />} label="Voted" />
-              </div>
-            )}
+              )}
```
andy-hook, line 92: Just checking, are we keeping the "You voted" tag? I might be wrong but I thought this app badge might be replacing it.

src/components/Proposals/TargetAppBadge.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { AppBadge, GU } from '@aragon/ui'
+import { useDescribeVote } from '../../hooks/useDescribeVote'
+import { useOrgApps } from '../../providers/OrgApps'
+import { getAppPresentation } from '../../utils/app-utils'
+import LoadingSkeleton from '../Loading/LoadingSkeleton'
+
+function TargetApp({ vote }) {
```
andy-hook, line 9: Let's give this component the same name `TargetAppBadge`

src/components/Proposals/TargetAppBadge.js
```diff
@@ -0,0 +1,78 @@
...
+  const {
+    emptyScript,
+    targetApp,
+    loading: descriptionLoading,
```
andy-hook, line 13: We can probably just go with `loading` in this new context, it's fairly obvious now what the loading refers to :)

src/components/Proposals/TargetAppBadge.js
```diff
@@ -0,0 +1,78 @@
...
+  } = useDescribeVote(vote.script, vote.id)
```
andy-hook, line 14: We should generally try to only pass the props needed by a component, I think here is a good chance to pass just `voteScript` and `voteId` as props.

andy-hook (approved):
Couple of small things we can improve now that it has its own component but otherwise LGTM!

src/components/Proposals/ProposalDetails/ProposalDetails.js
delfipolito: Hey, no. They both should be there. One is the target app and the other is the tag that appears when you already voted in the vote. :) like this voting app: 

image

PR #117, opened 2020-09-14
Settlement period ending time & dispute fixes

delfipolito:
- Settlement period: end date fix
- "Drafting jury" quote if vote status is disputed, and "Rulling executed" otherwise
- Dispute status info  change on colors 

image

### DIFF
--- a/src/components/Proposals/ProposalDetails/DisputableActionStatus.js
+++ b/src/components/Proposals/ProposalDetails/DisputableActionStatus.js
@@ -14,12 +14,12 @@ import { networkEnvironment } from '../../../current-environment'
 
 function DisputableActionStatus({ vote }) {
   const theme = useTheme()
-  console.log(vote)
   const disputableStatus = DISPUTABLE_VOTE_STATUSES.get(vote.status)
   const challenged = disputableStatus === VOTE_STATUS_CHALLENGED
   const scheduled = disputableStatus === VOTE_STATUS_SCHEDULED
   const challengeEndDate = toMs(vote.challengeEndDate)
   const pausedAt = toMs(vote.pausedAt)
+  const settledAt = toMs(vote.settledAt)
   const voteEndDate = toMs(vote.endDate)
   const extendedPeriod = toMs(vote.currentQuietEndingExtensionDuration)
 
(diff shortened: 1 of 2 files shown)


src/components/Proposals/ProposalDetails/DisputableActionStatus.js
```diff
@@ -34,7 +36,11 @@ function DisputableActionStatus({ vote }) {
         {challengeEndDate !== 0 && (
           <Item heading="Settlement period">
             <DisputablePeriod
-              endDate={challengeEndDate}
+              endDate={
+                toMs(vote.settledAt) > 0
+                  ? toMs(vote.settledAt)
+                  : challengeEndDate
```
andy-hook, line 42: For consistency let's format `settledAt` before using it in the template

src/components/Proposals/ProposalDetails/StatusInfo.js
```diff
@@ -14,7 +14,7 @@ function StatusInfo({ vote }) {
 
   const infoDescription = {
     [VOTE_STATUS_DISPUTED]: (
-      <InfoBox>
+      <InfoBox background="#FFEAEA" borderColor="#FF7C7C" color="#FF7C7C">
```
andy-hook, line 17: Can we use the `error` mode instead?

src/components/Proposals/ProposalDetails/DisputableActionStatus.js
```diff
@@ -13,6 +14,7 @@ import { networkEnvironment } from '../../../current-environment'
 
 function DisputableActionStatus({ vote }) {
   const theme = useTheme()
+  console.log(vote)
```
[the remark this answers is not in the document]
delfipolito: ohh, i didnt suppose to push this haha

src/components/Proposals/ProposalDetails/StatusInfo.js
delfipolito: Mmmhh, I dont think so. Error colors are diifferent. These are the ones that Paty handled me to put here

andy-hook: We could probably adopt the same colours for the errors in that case, but don't worry about that for now.

How about we use `css={`...`}` for these instead?

andy-hook (approved):

PR #126, opened 2020-09-17
Agreement: Subscribe to updates

andy-hook:
We're keeping votes and agreement subscriptions open in contexts as these routes are either frequented, or we will often need to consume this data in multiple places.


src/providers/AgreementSubscription.js
```diff
@@ -0,0 +1,172 @@
...
+  )
+  const [stakingFactory, stakingFactoryStatus] = useAgreement(
+    agreementApp,
+    (app) => app.stakingFactory()
+  )
+  const [signer, signerStatus] = useAgreement(
+    agreementApp,
+    (app) => (account ? app.onSigner(account) : null),
+    [account]
+  )
```
andy-hook, line 34: This code is very verbose but required because of the way the react library currently works. It's something I've been discussing with Pierre (among other things related to react api) and as we zone in on the refined next version we can come back here to tidy it up.

src/providers/AgreementSubscription.js
```diff
@@ -0,0 +1,172 @@
...
+    stakingFactoryStatus.loading ||
+    appsWithRequirementsLoading
+
+  const error =
+    agreementAppStatus.error ||
+    currentVersionStatus.error ||
+    disputableAppsStatus.error ||
+    signerStatus.error ||
+    stakingFactoryStatus.error ||
+    appsWithRequirementsError
```
andy-hook, line 59: Same as previous comment, this is ugly but readable, and a side affect of having so many separate statuses.

src/providers/AgreementSubscription.js
```diff
@@ -0,0 +1,172 @@
...
+    /* eslint-disable react-hooks/exhaustive-deps */
+  }, [
+    currentVersionUpdateValue,
+    appsWithRequirements,
+    stakingFactory,
+    signerUpdateValue,
+    loading,
+    error,
+  ])
+  /* eslint-enable react-hooks/exhaustive-deps */
```
andy-hook, line 86: Workaround for a memoization issue in the values returned from connect hook .

PR #132, opened 2020-09-21
Proposal: Subscribe to vote updates

andy-hook:
I had to work around a lack of subscription methods on `vote` as well as some known issues with the React api. This could definitely be improved but it does work consistently for the time being.


src/hooks/useVote.js
```diff
@@ -0,0 +1,192 @@
...
+
+const useDisputableVoting = createAppHook(
+  connectVoting,
+  connectorConfig.disputableVoting
+)
+
+export function useVote(proposalId) {
+  const [vote, loading] = useVoteSubscription(proposalId)
+  return [vote, loading]
+}
```
andy-hook, line 19: The intention is to do common unit formatting here, `useVoteSubscription` should be providing the original values.

src/components/Proposals/ProposalDetails/FeedbackModule.js
```diff
@@ -69,9 +69,9 @@ function FeedbackModule({ vote, mode }) {
                 <Strong>
                   {formatTokenAmount(
                     collateral.challengeAmount,
-                    collateralToken.decimals
+                    collateral.token.decimals
                   )}{' '}
-                  {collateralToken.symbol}
+                  {collateral.token.symbol}
```
andy-hook, line 74: I moved the collateral token inside the `collateral` object, I think it's clearer where to find it there.

src/hooks/useVote.js
```diff
@@ -0,0 +1,192 @@
...
+    () => (vote && account ? vote.onCastVote(account) : null),
+    [account, voteUpdateValue]
+  )
+
+  // Use caseVote as a dependency value for storing the latest vote
+  const castVoteUpdateValue = JSON.stringify(castVote)
+
+  /* eslint-disable react-hooks/exhaustive-deps */
+  const latestVote = useMemo(() => vote, [castVoteUpdateValue])
+  /* eslint-enable react-hooks/exhaustive-deps */
```
andy-hook, line 54: This is pretty awful in general, it basically uses the `onCastVote` subscription to decide when to refresh the `vote` object and trigger dependencies.

Ideally we would want to be subscribing to each value individually and only perform the required work based on what has changed, this isn't really possible to do at the moment though.

PR #135, opened 2020-09-22
Staking screen

delfipolito:
image

image


src/components/Stake/StakingMovements.js
```diff
@@ -0,0 +1,141 @@
...
+          Completed: {
+            background: theme.positiveSurface,
+            color: theme.positiveSurfaceContent,
+            icon: <IconCheck size="small" />,
+          },
+          Cancelled: {
+            background: theme.surfaceUnder,
+            color: theme.contentSecondary,
+            icon: <IconCross size="small" />,
+          },
```
andy-hook, line 59: I think we should generalise our [DisputableStatusLabel](https://github.com/aragon/network-dashboard/blob/master/src/components/Proposals/DisputableStatusLabel.js) to cover these cases. This will require a few changes elsewhere so we can do it in a separate improvement.

For some safety we should consider using `Symbols` mapped to these values in a similar way to how we manage our disputableStatuses.

src/components/Stake/StakingMovements.js
```diff
@@ -0,0 +1,141 @@
...
+            color: theme.surfaceOpened,
+            icon: (
+              <IconLock
+                size="small"
+                css={`
+                  margin-right: ${1 * GU}px;
+                `}
+              />
+            ),
+          },
```
andy-hook, line 83: To avoid the duplication we can apply the icon size and margin at the point where we render, e.g.

```javascript
Locked: {
    color: theme.surfaceOpened,
    icon: IconLock,
},
```

```javascript
{icon && <Icon size="small" css={...}}
```

src/components/Stake/StakingMovements.js
```diff
@@ -0,0 +1,141 @@
...
+            color: theme.negative,
+          },
+        }
+        return [
+          <time
+            dateTime={dateFormat(toMs(createdAt), 'standard')}
+            title={dateFormat(toMs(createdAt), 'standard')}
+          >
+            {dateFormat(toMs(createdAt), 'onlyDate')}
+          </time>,
```
andy-hook, line 97: Nice use of the `time` element!

src/components/Stake/StakingMovements.js
```diff
@@ -0,0 +1,141 @@
...
+      entries={stakingMovements}
+      renderEntry={({
+        amount,
+        createdAt,
+        actionState,
+        collateralState,
+        tokenDecimals,
+        disputableActionId,
+      }) => {
+        const actionAttributes = {
```
andy-hook, line 39: We should `useMemo` these objects outside of the render.

src/hooks/useStaking.js
```diff
@@ -0,0 +1,67 @@
...
+
+export function useStaking() {
+  const mounted = useMounted()
+  const { account } = useWallet()
+  const { agreementApp } = useOrgApps()
+  const [stakeManagement, setStakeManagement] = useState(null)
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    async function getStakingInformation() {
```
andy-hook, line 14: Let's chat about what to do with this hook :) we need to be subscribing and formatting the data and I think there's enough overlap to have it in the `agreementSubscription` and `useAgreement`. I can help with this!

src/hooks/useStaking.js
```diff
@@ -0,0 +1,67 @@
...
+      const defaultValues = {
+        available: '0',
+        challenged: '0',
+        locked: '0',
+        tokenDecimals: 1,
+        total: '0',
+      }
+      try {
+        if (account) {
+          const disputableApps = await agreementApp.disputableApps()
```
andy-hook, line 24: I made a change in another PR which broke the way `agreementApp` works (sorry!)

But if you want to get this working again for testing you can change `agreementApp` to `connectedAgreementApp`

src/providers/Staking.js
```diff
@@ -0,0 +1,36 @@
...
+
+const StakingContext = React.createContext({
+  staking: {},
+  loading: true,
+})
+
+function StakingProvider({ children }) {
+  const [stakeManagement, loading] = useStaking()
+
+  const Staking = useMemo(
```
andy-hook, line 13: As per previous comment, let's chat about this as we can probably use the existing `agreementSubscriptionProvider` for this :)

andy-hook (commented):
This is looking really great! my main points are about reworking our existing status labels to support additional states and looking at how we're getting data with the hook, the rest should be straightforward to address.

Let me know what you think, happy to work on the subscriptions with you and get that all sorted!

src/components/Stake/BalanceCard.js
```diff
@@ -28,14 +37,14 @@ function BalanceCard() {
           ${textStyle('title4')};
         `}
       >
-        17,746.21
+        {formatTokenAmount(total, tokenDecimals)}
```
andy-hook, line 40: Should we be be presenting with trailing zeros on even values?

src/components/InternalLink.js
```diff
@@ -0,0 +1,25 @@
...
+  const handlePageRequest = useCallback(() => {
+    history.push(to)
+  }, [history, to])
+
+  return (
+    <Link onClick={handlePageRequest} {...props}>
+      {children}
+    </Link>
+  )
+}
```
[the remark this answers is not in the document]
delfipolito: yes! :) I was planning on do it in the next PR

src/components/Stake/Stake.js
```diff
@@ -1,19 +1,57 @@
...
+
+  if (!stakeManagement) {
+    return (
+      <LayoutGutter>
+        <LayoutLimiter>
+          <Header primary="Stake management" />
+          <EmptyState
+            icon={noAccount}
+            title="Enable your account"
+            paragraph="Connect to an Ethereum provider to access your staking data. You may be temporarily redirected to a new screen."
```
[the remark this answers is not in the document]
delfipolito: To what place? I think this is making referrence to the account module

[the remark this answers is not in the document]
andy-hook: I figured it was referencing the provider explainer link (https://ethereum.org/en/wallets/) but it could also be an action to open the account module. Worth checking but probably not a big deal right now.

src/components/Stake/BalanceCard.js
delfipolito: yeah, if they are zeros, formatTokenAmount doesnt display them. But we can submit a PR on aragonUI too change this :)

andy-hook (approved):
LGTM! 🙌

